PR #401, opened 2017-10-18
Create CODE_OF_CONDUCT.md

omnivion:
GitHub wants a code of conduct for https://github.com/CMSgov/qpp-conversion-tool/community

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,1 @@
+1. Use common sense


CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1 @@
+1. Use common sense
```
halprin: I like this rule.

sfradkin: I'll follow up with @lucasmbrown-usds to see if there is a standard code of conduct that we should use here.

sfradkin:
Per @lucasmbrown-usds we will not add a Code of Conduct at this time.

PR #407, opened 2017-10-19
Error messages compiled

saquino0827:
### Information
- QPPCT-423

- compiles a list of all converter error messages into ERROR_MESSAGES.md

### Checklist 


halprin:
This list will go out of date as soon as we change one error message.  Can we, instead, make a `ErrorMessages.java` somewhere in our project that is an enum of all our error messages?  It could contain an error code that will help with @omnivion's QPPCT-424 story, be referenced where the error message is needed, and we can point whoever is interested in our error messages to this file.

For example,
```java
package gov.cms.qpp.conversion.model.error;

public enum ErrorMessages {

	//List of errors that output to users upon decoding and encoding failures
	NOT_VALID_XML(509, "The file is not a valid XML document"),
	//...

	//List of errors that output to users upon validation failure
	DENOM_NO_COUNT(341, "This Denominator Node does not have an Aggregate Count Node"),
	//...

	SOMETHING_ELSE(26, "The DogCow called {0} goes Moof!");

	private int errorCode;

	private String errorMessage;

	ErrorMessages(final int errorCode, final String errorMessage) {
		this.errorCode = errorCode;
		this.errorMessage = errorMessage;
	}

	public int getErrorCode() {
		return errorCode;
	}

	public String getErrorMessage() {
		return errorMessage;
	}

	@Override
	public String toString() {
		return getErrorMessage();
	}
}
```

sfradkin:
@halprin That's kind of what I was hoping would be created as part of QPPCT-424.  This is a stopgap for now so that we can have the error messages in an easy to read format because the users that will be evaluating the messages aren't necessarily going to want to look at actual source code.  We can certainly generate a file like this from an enum class though.

saquino0827:
I do like the idea of compiling this into a class and generating a md file from that class.

sfradkin:
I modified QPPCT-429 to specify autogenerating this file based on the work done in QPPCT-424.

PR #408, opened 2017-10-20
Do a clean when Dockerfiles make the application

halprin:
### Information
- Non-story nor issue improvement.

- Initiate a clean when building our application in Docker.  This will delete any extraneous files when we copy our application into Docker.
- Saves about 100 MB when sending our Docker image to our small Flexion endpoint.

### Checklist 

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,7 +7,7 @@ COPY ./ /usr/src/app/
 
 WORKDIR /usr/src/app/
 
-RUN mvn install -Dmaven.test.skip -Djacoco.skip=true > /dev/null
+RUN mvn clean install -Dmaven.test.skip -Djacoco.skip=true > /dev/null
 RUN cp ./rest-api/target/rest-api.jar /usr/src/run/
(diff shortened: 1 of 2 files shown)


omnivion:
Well this isn't actually that helpful. From what I know of maven clean plugin, it cleans the directory BEFORE the build, not after, so any file pollution produced by the build will still be there. Does this solve another problem?

halprin:
You are correct that it runs before the build.  This really only applies to the docker build that we run when we do in CircleCI.  In the CircleCI case, we do
1. Our usual `mvn clean verify` as specified in `circle.yml`.
1. Copy all that created stuff into the container as part of the docker build.
1. And finally we do the `mvn clean install` in the docker container.

There could be additional things left over from step 1 that gets copied over in step 2.  That is why in step 3, I added `clean`.

Hmmm, perhaps if I add `mvn clean` once we grab the ReST API after the build...

halprin:
Don't merge this yet.  I still want to tinker with this PR some more.

halprin:
Don't have time to dedicate to this PR at this time.  I'll reopen it when I have time to tinker with this some more.

PR #411, opened 2017-10-24
QPPCT-407: circleci v2

clydet:
### Information
- JIRA story QPPCT-407

- bump circleci to V2

### Checklist 


sfradkin:
@clydet: I assigned QPPCT-320 to you since we had that already for converting to CircleCI 2

.circleci/config.yml
```diff
@@ -0,0 +1,40 @@
...
+            cp -R rest-api/target/site/jacoco $ARTIFACT_DIR/coverage/commandline;
+
+      - store_artifacts:
+          path: /tmp/artifacts
+
+workflows:
+  version: 2
+  build_backend:
+    jobs:
+      - build
```
sfradkin: I assume that there will be another PR at some point for 407 to add a job definition for the front end build?

clydet: Yes, digging up some history on the simple ui build

.circleci/config.yml
```diff
@@ -0,0 +1,40 @@
...
+    working_directory: ~/qpp-conversion-tool
+    docker:
+      - image: circleci/openjdk:8-jdk-browsers
+    steps:
+      - checkout
+      - run: cd converter; mvn dependency:resolve
+      - run: wget -O sonarscanner.zip https://sonarsource.bintray.com/Distribution/sonar-scanner-cli/sonar-scanner-cli-3.0.1.733.zip
+      - run: unzip -d . sonarscanner.zip
+      - run: wget -O get-pip.py https://bootstrap.pypa.io/get-pip.py
+      - run: sudo python get-pip.py
```
halprin: There's no way to specify an environment that has Python and PIP already installed in CircleCI 2.0?  We need to do this manually?

clydet: I still had to install pip when I used the python container as well. It had both python 2 and 3 installed, but for whatever reason neither came with pip.

PR #416, opened 2017-10-27
QPPCT-378: Load Test IMPL Environment

halprin:
### Information
- JIRA story QPPCT-378.

- Copied QRDA3 sample files into the `small`, `standard`, and `large` based on their size.
- Renamed JMeter test plans based on the order we want them executed.
- Created a new JMeter test that ramps up from 1 request per second to `rps_micro` (45 currently) over one minute and then stays at `rps_micro` for the following 9 minutes.

### Checklist 


benchmark/pom.xml
```diff
@@ -135,8 +135,8 @@
 						<file_large>large</file_large>
 					</propertiesJMeter>
 					<jMeterProcessJVMSettings>
-						<xms>4048</xms>
-						<xmx>4048</xmx>
+						<xms>12288</xms>
+						<xmx>12288</xmx>
```
clydet: beasty

halprin:
@clydet The QRDA3 files in this PR are copied from the `sample-files` directory.  The problem is that I need to isolate the small, medium, and large files from each other for the different tests plans.  Perhaps we can do the small, medium, and large directories in `sample-files`, but then that seems unnecessary for humans just looking for some sample files.

PR #421, opened 2017-10-31
QPPCT-452 : Fix case insensitivity with all ecqm UUID's

saquino0827:
### Information
- Case Insensitivity validation with eCQM Uuid's.

- Forces all eCQM Measure Uuid's to be decoded into a lowercase format to match the measure-data.json. This is because we use a map to compare values which enforces the lowercase formatting.
- Refactored the MIPS and CPC+ measure data uuid checks to be equalsIgnoreCase as strings.
- Added roundtrip for coverage.

### Checklist 


converter/src/main/java/gov/cms/qpp/conversion/validate/QualityMeasureIdValidator.java
```diff
@@ -305,7 +305,7 @@ protected void addMeasureConfigurationValidationMessage(Supplier<Object> check,
 			thoroughlyCheck(thisNode)
 					.incompleteValidation()
 					.singleValue(message, name);
-			return uuid.get().equals(thisNode.getValue(name));
+			return uuid.get().toString().equalsIgnoreCase(thisNode.getValue(name));
```
omnivion: why toString()

saquino0827: You're right. Changing that.

converter/src/main/java/gov/cms/qpp/conversion/validate/MipsQualityMeasureIdValidator.java
```diff
@@ -121,7 +121,7 @@ private void validatePerformanceRateUuidExists(Node performanceRateNode) {
 	 * @return
 	 */
 	private Predicate<SubPopulation> makePerformanceRateUuidFinder(String uuid) {
-		return subPopulation -> uuid.equals(subPopulation.getNumeratorUuid());
+		return subPopulation -> uuid.toString().equalsIgnoreCase(subPopulation.getNumeratorUuid());
```
sfradkin: @saquino0827: Please change this also

converter/src/main/java/gov/cms/qpp/conversion/validate/QualityMeasureIdValidator.java
saquino0827: Oh I was looking at the wrong part of my code. The reason for that is because uuid.get() doesn't register as a string and needs to be a string in order to use the `equalsIgnoreCase` operation. I did change another piece of code to not use toString from this so thank you.

converter/src/main/java/gov/cms/qpp/conversion/validate/MipsQualityMeasureIdValidator.java
saquino0827: Yes this is the one I changed just now.

converter/src/main/java/gov/cms/qpp/conversion/validate/QualityMeasureIdValidator.java
saquino0827: The get() method from Supplier returns T and doesn't know it's a string. The only other way I could thing of is casting it as a string. Which could be better.

saquino0827: I am trying an approach of changing the `Supplier` to be `<String>` instead of `<Object>`. We'll see how this fairs.

saquino0827: Alright. Changed and let me know if it's better!... or worse. Thanks for pointing that out.

halprin: Nice fix!

converter/src/main/java/gov/cms/qpp/conversion/decode/QualityMeasureIdDecoder.java
```diff
@@ -39,7 +39,7 @@ protected DecodeResult internalDecode(Element element, Node thisNode) {
 		List<String> measureGuids = getMeasureGuid(element);
 
 		measureGuids.forEach(measureGuid ->
-			thisNode.putValue(MEASURE_ID, measureGuid, false));
+			thisNode.putValue(MEASURE_ID, measureGuid.toLowerCase(), false));
```
FlexionSonarQube: MINOR Define the locale to be used in this String operation. [rule](https://sonarcloud.io/coding_rules#rule_key=squid%3AS1449)

FlexionSonarQube:
SonarQube analysis reported 1 issue
* ![MINOR][MINOR] 1 minor

Watch the comments in this conversation to review them.

[MINOR]: https://sonarsource.github.io/sonar-github/severity-minor.png 'Severity: MINOR'

converter/src/test/java/gov/cms/qpp/acceptance/QualityMeasureIdRoundTripTest.java
```diff
@@ -76,4 +79,19 @@ public void testMeasureCMS160v5PerformanceRateUuid() throws IOException {
...
+	@Test
+	public void testMeasureCMS52v5WithInsensitiveTextUuid() throws IOException {
+		Converter converter = new Converter(new PathQrdaSource(INSENSITIVE_TEXT_FILE));
+		List<Detail> details = new ArrayList<>();
+
+		try {
+			converter.transform();
+		} catch (TransformException exception) {
+			AllErrors errors = exception.getDetails();
+			details.addAll(errors.getErrors().get(0).getDetails());
```
halprin: I would prefer to make this test a bit more specific.  Can we use the special Google Truth thing we created to test validation errors to test that we do _not_ get a specific error that would occur if we accidentally became case sensitive again?

converter/src/main/java/gov/cms/qpp/conversion/decode/QualityMeasureIdDecoder.java
clydet: What's the reason for the case change if we're ignoring case when performing the comparison?

converter/src/test/java/gov/cms/qpp/acceptance/QualityMeasureIdRoundTripTest.java
saquino0827: I refactored this test to specifically assert for the missing `Electronic measure uuid` message.
I added another test to assert for measure data uuid's being incorrect as well.

converter/src/main/java/gov/cms/qpp/conversion/decode/QualityMeasureIdDecoder.java
saquino0827: The reason is because we perform a map check in `QualityMeasureIdValidtor.java` based on this value compared to the `measureData.json` if the eMeasureId doesn't completely match, it doesn't find the correct Uuid to validate. The ignore case validation is strictly for the measure data's sub population uuid's i.e. DENOM, DENEX, IPOP, etc...

saquino0827: You can see the validation in the method `QualityMeasureIdValidator.validateMeasureConfigs`

PR #422, opened 2017-11-01
QPPCT-442: time boxed junit5 conversion

clydet:
### Information
- Fixes sonarqube updates
- JIRA QPPCT-442

- Fixes publishing of sonar results
- some junit 5 migration including a MockitoExtension

### Checklist 


tools/bin/sonarqube.sh
```diff
@@ -4,7 +4,7 @@ DIR=$(dirname "$0")
 SONAR_HOST=https://sonarcloud.io
 ORG_KEY=cmsgov
 
-if [[ "$CIRCLE_BRANCH" == "master" || ( ! -z $SONAR_OTHER_BRANCH && "$CIRCLE_BRANCH" == "$SONAR_OTHER_BRANCH" ) ]]; then
+#if [[ "$CIRCLE_BRANCH" == "master" || ( ! -z $SONAR_OTHER_BRANCH && "$CIRCLE_BRANCH" == "$SONAR_OTHER_BRANCH" ) ]]; then
```
halprin: This prevents the preview from running and getting feedback from SonarQube in our PR and instead does a full scan with the results being uploaded to our SonarQube project.  Is that what you meant to happen?

sfradkin: I was just going to ask that.  I'm not sure that's a bad thing, necessarily.  However, we should maybe limit this to just PRs otherwise we'll get pretty wild results over on the sonarcloud website.

halprin: The problem with doing a full scan on a PR is that it is harder to see the results on the full SonarQube server.  For example, you'll need to somehow filter to your code changes only.  Also, you'll need to look up your results before another PR build because I think we loose the previous results.  Does anyone know of a way to go back to see the results of a scan from a previous time?

tools/bin/sonarqube.sh
```diff
@@ -19,25 +19,25 @@ if [[ "$CIRCLE_BRANCH" == "master" || ( ! -z $SONAR_OTHER_BRANCH && "$CIRCLE_BRA
...
-		                                            -Dsonar.github.pullRequest=${PR_NUMBER} \
-		                                            -Dsonar.github.repository=CMSgov/qpp-conversion-tool \
-		                                            -Dsonar.github.oauth=${SONAR_PR_KEY} \
-													-Dsonar.organization=${ORG_KEY} \
-		                                            -Dsonar.login=${SONAR_KEY_NEW}
-	done
-else
-	echo "Not on master nor in a pull request."
-fi
+#elif [[ ! -z $CI_PULL_REQUESTS ]]; then
```
saquino0827: Are these commented out for a reason? Or can we remove them.

rest-api/src/test/java/gov/cms/qpp/conversion/api/services/ValidationServiceImplTest.java
```diff
@@ -44,7 +42,8 @@
 import static org.mockito.Mockito.verifyZeroInteractions;
 import static org.mockito.Mockito.when;
 
-@RunWith(MockitoJUnitRunner.class)
+
+@ExtendWith(MockitoExtension.class)
 public class ValidationServiceImplTest {
```
omnivion: class doesn't need to be public

clydet: yeah, got that with the commit below

tools/bin/sonarqube.sh
clydet: I wanted to ensure that the publishing of sonar results was functioning, so I commented the master branch condition

tools/bin/sonarqube.sh
clydet: I was testing the publish, removed the comments

.circleci/config.yml
```diff
@@ -14,6 +14,8 @@ jobs:
       - run: sudo pip install ansible
       - run: cp ./sonar-scanner.properties sonar-scanner-3.0.1.733/conf/sonar-scanner.properties
       - run: mvn clean verify
+      - run: curl -sL https://deb.nodesource.com/setup_8.x | sudo -E bash -
+      - run: sudo apt-get install -y nodejs
```
clydet: I'm working on another change set that'll introduce a custom build image so as to remove a lot of the dependency installation from the build job configurations.

sfradkin:
@clydet Can you remove the demo files from this PR and add them to the deploy repo?

rest-api/src/test/java/gov/cms/qpp/conversion/api/services/AnyOrderAsyncActionServiceTest.java
```diff
@@ -133,21 +133,13 @@ public void testDependencyOrder() {
...
-		try {
-			completableFuture.join();
-			fail("A CompletionException was not thrown.");
-		} catch (CompletionException exception) {
-			assertThat(exception).hasCauseThat().isInstanceOf(UncheckedInterruptedException.class);
-			assertThat(objectUnderTest.timesAsynchronousActionCalled.get()).isEqualTo(1);  //not two
-		} catch (Exception exception) {
-			fail("A CompletionException was not thrown.");
-		}
+		CompletionException exception = assertThrows(CompletionException.class, completableFuture::join);
```
halprin: This `assertThrows` is quite nice!

PR #434, opened 2017-11-03
Fix/error file

saquino0827:
### Information
- Fixes error_messages.md file to have updated error codes.


.circleci/config.yml
```diff
@@ -82,7 +82,7 @@ workflows:
       - ui_deployment:
           filters:
             branches:
-              only: master
+              only: ui_deployment
```
saquino0827: I do not know why this is listed as a change. It's already in master.

PR #449, opened 2017-11-14
Dependency update

omnivion:
### Information
- JIRA story QPPCT-492.

- Updates most of our dependencies

### Checklist 


rest-api/pom.xml
```diff
@@ -106,21 +106,21 @@
...
-			<version>0.1.4</version>
+			<version>0.1.7</version>
 			<scope>test</scope>
 		</dependency>
 
 		<dependency>
 			<groupId>org.springframework</groupId>
 			<artifactId>spring-test</artifactId>
-			<version>4.3.10.RELEASE</version>
+			<version>5.0.1.RELEASE</version>
```
halprin: This is a big jump here.  Plus, SpringBoot version 1.5.8 is still on the Spring 4.* release.  I would keep this at the latest and greatest Spring 4.* release.

omnivion: It's just the testing framework, and all tests pass.

sfradkin: It doesn't try to bring in the Spring 5.0 libraries with this?

omnivion: There's an <exclusions> block which excludes *.

PR #451, opened 2017-11-15
Remove extraneous {} from conversion error

halprin:
### Information
An enhancement.

- Removed extraneous `{}` since it wasn't being used.

### Checklist 

### DIFF
--- a/commandline/src/main/java/gov/cms/qpp/conversion/ConversionFileWriterWrapper.java
+++ b/commandline/src/main/java/gov/cms/qpp/conversion/ConversionFileWriterWrapper.java
@@ -1,20 +1,19 @@
 package gov.cms.qpp.conversion;
 
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.ObjectWriter;
+import gov.cms.qpp.conversion.encode.JsonWrapper;
+import gov.cms.qpp.conversion.model.error.AllErrors;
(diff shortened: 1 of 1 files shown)


commandline/src/main/java/gov/cms/qpp/conversion/ConversionFileWriterWrapper.java
```diff
@@ -67,7 +66,7 @@ private void executeConverter(Converter converter) {
 		} catch (TransformException exception) {
 			AllErrors allErrors = exception.getDetails();
 			Path outFile = getOutputFile(source.getName(), false);
-			DEV_LOG.warn("There were errors during conversion.  Writing out errors to {} " + outFile.toString(),
+			DEV_LOG.warn("There were errors during conversion.  Writing out errors to " + outFile.toString(),
```
sfradkin: You can probably just leave the `{}` in there and change it to `DEV_LOG.warn("There ... to {}", outfile.toString(), exception);`

I believe this was preferred over appending.

halprin: Sadly no, this does not work.

The Logger only spits out the exception details if the method `Logger#warn(String msg, Throwable t)` is called.

If you call `Logger#warn(String format, Object... arguments)`, then the array of `Object`s are assumed to be things to fill in the format string.  I could do `DEV_LOG.warn("There ... to {} {}", outFile.toString(), exception)`, but many of the exception details are not displayed in that case.

PR #452, opened 2017-11-15
Modified and added more logging to the ReST API

halprin:
### Information
Improvement task.

- Modified existing logging statements to be more clear.
- Added new logging statements to make future triaging easier.

### Checklist 

### DIFF
--- a/rest-api/src/main/java/gov/cms/qpp/conversion/api/controllers/v1/QrdaControllerV1.java
+++ b/rest-api/src/main/java/gov/cms/qpp/conversion/api/controllers/v1/QrdaControllerV1.java
@@ -52,16 +52,17 @@ public class QrdaControllerV1 {
 	 */
 	@RequestMapping(method = RequestMethod.POST, headers = {"Accept=" + Constants.V1_API_ACCEPT})
 	public ResponseEntity<String> uploadQrdaFile(@RequestParam MultipartFile file) throws IOException {
-		API_LOG.info("Conversion request received");
+		String originalFilename = file.getOriginalFilename();
+		API_LOG.info("Conversion request received for " + originalFilename);
 
 		Converter.ConversionReport conversionReport = qrdaService.convertQrda3ToQpp(
-				new InputStreamSupplierQrdaSource(file.getName(), inputStreamSupplier(file)));
+				new InputStreamSupplierQrdaSource(originalFilename, inputStreamSupplier(file)));
 
 		validationService.validateQpp(conversionReport);
 
 		auditService.success(conversionReport);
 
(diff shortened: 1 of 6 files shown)


rest-api/src/main/java/gov/cms/qpp/conversion/api/controllers/v1/QrdaControllerV1.java
```diff
@@ -52,13 +52,17 @@
 	 */
 	@RequestMapping(method = RequestMethod.POST, headers = {"Accept=" + Constants.V1_API_ACCEPT})
 	public ResponseEntity<String> uploadQrdaFile(@RequestParam MultipartFile file) throws IOException {
-		API_LOG.info("Request received " + file.getName());
+		API_LOG.info("Conversion request received");
```
halprin: Authors note: `file.getName()` only ever returned `file`.  Not exactly helpful.  Thankfully, the log automatically appends a request ID during a request so we can track it.

sfradkin: So does this mean that we can't actually glean the original file name of the uploaded file?

halprin: Did some digging.  There is also `file.getOriginalFilename()` that does work.  I will update this PR to use that.

rest-api/src/main/java/gov/cms/qpp/conversion/api/controllers/v1/QrdaControllerV1.java
```diff
@@ -52,13 +52,18 @@
 	 */
 	@RequestMapping(method = RequestMethod.POST, headers = {"Accept=" + Constants.V1_API_ACCEPT})
 	public ResponseEntity<String> uploadQrdaFile(@RequestParam MultipartFile file) throws IOException {
-		API_LOG.info("Request received " + file.getName());
+		String originalFilename = file.getOriginalFilename();
+		API_LOG.info("Conversion request received for " + originalFilename);
+
 		Converter.ConversionReport conversionReport = qrdaService.convertQrda3ToQpp(
-				new InputStreamSupplierQrdaSource(file.getName(), inputStreamSupplier(file)));
+				new InputStreamSupplierQrdaSource(originalFilename, inputStreamSupplier(file)));
```
halprin: This will reflect in our DynamoDB audit table now too.  No more boring `file`.

sfradkin: excellent!

PR #463, opened 2017-11-21
QPPCT-494: set base href

clydet:
### Information
- JIRA QPPCT-494

- resolve references to bundles

### Checklist 


halprin:
This will make us dependent on the AdHoc API gateway.  I guess we're OK with that to make it work without the ending slash?

clydet:
@halprin suppose, but in and of itself won't impact local usage.

PR #465, opened 2017-11-25
QPPCT-509: regard filename as pii

clydet:
### Information
- JIRA QPPCT-509

- regard filename as PII

### Checklist 


sfradkin:
Just want to verify that a search was made in the codebase for anything filename related.

Also, if there is a way to perform an integration test to verify that the filename value is being encrypted, it would be good to have in here.

saquino0827:
An integration test @sfradkin seems like a good idea for how serious the PII issue was. Are we adding that in this PR? Otherwise, I'll approve.

converter/src/main/java/gov/cms/qpp/conversion/Converter.java
```diff
@@ -80,7 +80,7 @@ public Context getContext() {
 	 * @return status of conversion
 	 */
 	public JsonWrapper transform() {
-		DEV_LOG.info("Transform invoked with file {}", source.getName());
```
omnivion: I think we should document why we're making this change in the code

clydet:
@halprin true, there is an occurrence in command line, but that's not of concern.

clydet:
@sfradkin not sure how feasible that will be. The file name will be encrypted at rest so before write and after read it will be decrypted.

PR #471, opened 2017-11-29
Run filesystem tests against unix, mac, and windows

omnivion:
### Information
- Fixes QPPCT-521.

### Checklist 


test-commons/src/main/java/gov/cms/qpp/test/jimfs/FileTestHelper.java
```diff
@@ -11,10 +11,12 @@
...
 import com.google.common.jimfs.Jimfs;
 
+import gov.cms.qpp.test.CopyFileVisitor;
+
 public class FileTestHelper {
 
-	public static FileSystem createMockFileSystem() {
-		FileSystem mock = Jimfs.newFileSystem(Configuration.unix());
+	public static FileSystem createMockFileSystem(Configuration configuration) {
+		FileSystem mock = Jimfs.newFileSystem();
```
saquino0827: I'm a little confused why we pass in configuration and don't use it.

omnivion: good catch!

PR #477, opened 2017-12-04
QPPCT-533

sfradkin:
adding sample files for CPC+ specific validations and documenting in README

### Information
- JIRA story QPPCT-533

- Adds new CPC+ sample files
- Updates README
-

### Checklist 
- [n/a] All JUnit tests pass (`mvn clean verify`).


clydet:
@sfradkin do we want to fold these in with the existing CPC regression scenarios?

sfradkin:
@clydet These are all built off of the existing CPC+ regression scenarios.  They cover the CPC+ specific validations we have implemented in order to give a range of errors.

bijujoseph:
Thank you @sfradkin. cc: @sietekk

PR #478, opened 2017-12-05
QPPCT-511: use large instances to run build jobs

clydet:
### Information
- JIRA QPPCT-511

- Specified the use of a larger instance or resource class on which our build jobs should be executed.

### Checklist 

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,6 +4,7 @@ jobs:
     working_directory: ~/qpp-conversion-tool
     docker:
       - image: circleci/openjdk:8-jdk-browsers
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -39,6 +40,7 @@ jobs:
     working_directory: ~/qpp-conversion-tool
     docker:
       - image: circleci/node:7.10.0-browsers
+    resource_class: large
```
halprin: On second thought, though, we don't really need the `large` resource class for `ui_test` and `ui_deployment` because we're only running into problems with `build`.

PR #491, opened 2017-12-08
Restrict memory usage with MAVEN_OPTS

omnivion:
### Information
- JIRA story QPPCT-536.

- Lowers memory dedicated to the build through MAVEN_OPTS on circleci

### Checklist 

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,6 +4,8 @@ jobs:
     working_directory: ~/qpp-conversion-tool
     docker:
       - image: circleci/openjdk:8-jdk-browsers
+    environment:
+      MAVEN_OPTS: "-Xms1G -Xmx1G"
     resource_class: large
     steps:
       - checkout


.circleci/config.yml
```diff
@@ -4,6 +4,8 @@ jobs:
     working_directory: ~/qpp-conversion-tool
     docker:
       - image: circleci/openjdk:8-jdk-browsers
+    environment:
+      MAVEN_OPTS: "-Xms1G -Xmx1G"
```
halprin: We've been allocated the large resource pool now which gives us 4 cores and 8192 MB RAM.  Perhaps set `Xmx` to `7G`?

omnivion: We can try larger amounts later - this passes just fine in pretty much the same time, and circle isn't killing it. Better to make sure all the other builds are passing first before we mess around with this further imo.

saquino0827: Agreed.

PR #504, opened 2017-12-12
Don't use logging to assert functionality

halprin:
### Information
- Enhancement PR

- The log capturing wasn't working in IntelliJ, so changed the assertions to assert on something other than logs.

### Checklist 


omnivion:
Weird, what was happening in intellij?

halprin:
@omnivion When running the unit tests in IntelliJ's built-in runner, it doesn't know that it should _not_ include logback.  So because logback is on the classpath, it wins against `uk.org.lidalia:slf4j-test`, and so the log capturing doesn't work.

PR #508, opened 2017-12-14
QPPCT-414: Reduced string creation rate

halprin:
### Information
- JIRA story QPPCT-414.

- While triaging calls to `new Reflections(...)` as per the QPPCT-414 description, I noticed that a lot of strings being creating at certain logger debug calls in high traffic methods.
- Due to the string concatenation in the logger debug statements, strings were being created even though debug logging was disabled.  Strings are now created only when debug logging is enabled.
- Deleted unused code from the debug logging refactor.
- Moved the memory expensive `XPathHelper#getAbsolutePath` call to below the `DecodeResult.TREE_ESCAPED` check.  This results in the call occurring only if the child node that was just decoded will stick around.

### Checklist 


halprin:
@omnivion You're right, 7 GB apparently is too close to the limit of the 8 GB.  I just had a `The forked VM terminated without properly saying goodbye. VM crash or System.exit called?`.  Sadness.

PR #509, opened 2017-12-15
QPPCT-495: Rename QrdaSource and add a size

halprin:
### Information
- JIRA story QPPCT-495.

- This is not the complete story.
- Tell me if I'm bananas where I'm going with this story.
- Added a `getSize` method to `QrdaSource` and added implementations.
  - This is needed to report the size to the AWS S3 SDK when we upload the file.
- Renamed the `QrdaSource` to `Source` and all implementing classes by removing the `Qrda` part.  Why?
  - There are other things we upload to S3 such as the QPP or any validation errors, and I was thinking that we could hold the `InputStream` and its size also in a `Source`.  That way we can report the size to the AWS S3 SDK for these items too.

### Checklist 


converter/src/test/java/gov/cms/qpp/conversion/ConverterTest.java
```diff
@@ -50,17 +50,17 @@
...
 		//no exception should be thrown, hence explicitly stating the expected exception is None
 	}
 
 	@Test(expected = org.junit.Test.None.class)
-	public void testValidQppStream() {
+	public void testValidQppStream() throws IOException {
 		Path path = Paths.get("../qrda-files/valid-QRDA-III-latest.xml");
 		Converter converter = new Converter(
-				new InputStreamSupplierQrdaSource(path.toString(), () -> XmlUtils.fileToStream(path)));
+				new InputStreamSupplierSource(path.toString(), () -> XmlUtils.fileToStream(path), Files.size(path)));
```
clydet: Is there a reason you favored this constructor over the 2 arg constructor?

halprin: Because the two argument constructor pulls the `InputStream` into memory to compute the size.  This three argument constructor takes the size and just uses it without having to pull the entire source into memory.

PR #514, opened 2017-12-19
QPPCT-552 run functional specs

clydet:
### Information
- QPPCT-552

- Better build job granularity
- Restore sandbox deployment
- Introduce api acceptance testing

### Checklist 


.circleci/config.yml
```diff
@@ -69,20 +111,35 @@ jobs:
...
-      - build
-      - ui_test
+      - checkout
+      - build:
+          requires:
+            - checkout
+      - ui_test:
+          requires:
+            - checkout
+      - deploy_api:
```
halprin: I'm hesitant to have `deploy_api` only depend on `checkout`.  While the docker image build does do its own independent build of our application, it does not do any tests.  So, this could result in `build` failing, but the `deploy_api` succeeding.  Is that OK?

rest-api/src/test/java/gov/cms/qpp/conversion/api/acceptance/HealthApiAcceptance.java
```diff
@@ -0,0 +1,17 @@
...
+
+import static io.restassured.RestAssured.get;
+
+
+@ExtendWith(RestExtension.class)
+class HealthApiAcceptance {
+	@Test
+	@Tag("acceptance")
+	void apiResponsesWithOkStatus() {
+		get("/health").then().statusCode(200);
```
halprin: Nice!  We should figure out what should be tested with one of these "Acceptance" test versus one of the "Integration" tests like `QrdaRestIntegration.java`.

.circleci/config.yml
clydet: As it stands unit test failures won't be of concern. This job will only run on merges to master.

rest-api/src/test/java/gov/cms/qpp/conversion/api/acceptance/HealthApiAcceptance.java
clydet: Agreed, we should add some tasking to the backlog to track scenarios that are currently absent or should be converted to acceptance tests.

PR #520, opened 2017-12-21
Use US/Eastern time instead of UTC

omnivion:
### Information
- JIRA story QPPCT-562.

- Use US/Eastern ZoneId'd clock rather than UTC clock for validation

### Checklist 

### DIFF
--- a/converter/src/main/java/gov/cms/qpp/conversion/validate/CpcClinicalDocumentValidator.java
+++ b/converter/src/main/java/gov/cms/qpp/conversion/validate/CpcClinicalDocumentValidator.java
@@ -2,6 +2,7 @@ package gov.cms.qpp.conversion.validate;
 
 import java.time.Clock;
 import java.time.LocalDate;
+import java.time.ZoneId;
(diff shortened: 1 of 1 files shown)


converter/src/main/java/gov/cms/qpp/conversion/validate/CpcClinicalDocumentValidator.java
```diff
@@ -24,8 +25,8 @@
 	static final String END_DATE_VARIABLE = "CPC_END_DATE";
 	private static final String NEVER_ENDING = "3000-01-01";
 	// LocalDate.now() creates extra unneeded clock objects before Java 9.
-	// It also uses the system clock, rather than UTC.
-	private static final Clock CLOCK = Clock.systemUTC();
+	// It also uses the system clock, rather than Eastern Time.
+	private static final Clock CLOCK = Clock.system(ZoneId.of("US/Eastern"));
```
clydet: test worthy?

PR #532, opened 2018-01-02
QPPCT-566: Restore integration tests

halprin:
### Information
- JIRA story QPPCT-566.

- `SubmissionIntegrationTest` has been fixed up and now runs as part of the build.
  - Removed submission delete code since the validation URL is being used.
- `QrdaRestIntegrationTest` has been fixed up and now runs as part of the build.
  - Had to revert the `spring-test` dependency back to 4.x because the 5.x didn't work with our 4.x Spring application.
  - That resulted in having to revert the JUnit test back to JUnit vintage (4).

### Checklist 


rest-api/src/test/java/gov/cms/qpp/conversion/api/integration/QrdaRestIntegrationTest.java
```diff
@@ -1,54 +1,46 @@
...
+import java.nio.file.Paths;
+
+import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.content;
+import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.jsonPath;
+import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.status;
 
 @SpringBootTest
 @WebAppConfiguration
-@ExtendWith(SpringExtension.class)
+@RunWith(SpringRunner.class)
```
omnivion: Any reason for this?

halprin: It extends from the fact that the `spring-test` dependency must be rolled back to 4.x.  If it isn't, these tests don't run correctly.  The 4.x version doesn't have support for JUnit 5.

rest-api/src/test/java/gov/cms/qpp/conversion/api/integration/HealthRestIntegrationTest.java
```diff
@@ -0,0 +1,48 @@
...
+
+	@Test
+	public void shouldBeHealthy() throws Exception {
+		mockMvc.perform(MockMvcRequestBuilders.request(HttpMethod.GET, "/health"))
+			.andExpect(status().is(200))
+			.andExpect(content().contentType(MediaType.APPLICATION_JSON_UTF8))
+			.andExpect(jsonPath("$.environmentVariables").exists())
+			.andExpect(jsonPath("$.systemProperties").exists())
+			.andExpect(jsonPath("$.implementationVersion").exists());
+	}
```
saquino0827: aw yeah 👍

PR #553, opened 2018-01-10
QPPCT-599, add debug to the maven compile config

sfradkin:
### Information
- Fixes QPPCT-599

- Adds <debug>true</debug> to the maven compiler configuration to add debug info to the compiled java classes
-
-

### Checklist 


sfradkin:
After further reading, it appears that this won't get us deeper tracing in New Relic.

PR #567, opened 2018-01-15
Remove Fixes from PR template

omnivion:
- Updated the PR template

### Checklist 


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,5 +1,4 @@
 ### Information
-- Fixes #_.
```
halprin: Why are we removing this?

sfradkin: I would think that "Fixes" is somewhat redundant to "JIRA story".  I usually just put the JIRA number here and the actual changes in the next section.

omnivion: ^ what Scott said

saquino0827: I could see this as a repeat of the Jira ticket's title or description. I usually just delete the Fixes and add the ticket number. I would be further inclined to just change it to say something like `Fixes or JIRA Ticket`. The reason why Fixes is there is if there are any developers who aren't on our JIRA working on fixes for this project. But I like it better in one line.

halprin: This line was added originally for the users who create a PR that fixes a GitHub issue.  GitHub keys off `Fixes #23` which will close GitHub issue 23 as fixed once the PR is merged into `master`.

PR #573, opened 2018-01-17
HTTP method specific annotations

omnivion:
### Information
- JIRA story QPPCT-608.

### Checklist 


rest-api/src/main/java/gov/cms/qpp/conversion/api/controllers/HealthCheckController.java
```diff
@@ -34,7 +35,7 @@ public HealthCheckController(final VersionService version) {
 	 *
 	 * @return health check of version, environment variables, and system properties
 	 */
-	@RequestMapping(method = RequestMethod.GET)
+	@GetMapping
```
halprin: Oh, I thought you were talking about `@GET`, etc. originally.

omnivion: @halprin that might work too. Never tried it, cool idea. We can look into it later.

PR #587, opened 2018-01-24
QPPCT-613: wrapper metadata exclusion test

clydet:
### Information
- QPPCT-613

- added means for encodeLabel collision resolution for unnamed metadata holders

### Checklist 


converter/src/main/java/gov/cms/qpp/conversion/encode/AciSectionEncoder.java
```diff
@@ -119,7 +119,7 @@ private void encodeReportingParameter(JsonWrapper wrapper, Node node) {
 			DEV_LOG.error("Missing Reporting Parameters from ACI Section");
 			return;
 		}
-		reportingParamEncoder.encode(wrapper, reportingChild);
+		reportingParamEncoder.encode(wrapper, reportingChild, false);
```
halprin: Do you think we run into the same problem for the IA and Quality sections too?

clydet: Yes, those sections would also be vulnerable

halprin: Should we address them in this PR or a different PR?  I think these changes would be easy to extend to the other sections.

clydet: They both extend from AciSectionEncoder

halprin: Oh......

Not very bright

PR #590, opened 2018-01-25
A simple speed test to measure conversion time

halprin:
### Information
- Generic improvement.

- Just a simple speed test that averages results.

### Checklist 


benchmark/src/main/java/gov/cms/qpp/conversion/SpeedTest.java
```diff
@@ -0,0 +1,73 @@
...
+import com.google.common.collect.Lists;
+
+import java.nio.file.Paths;
+import java.util.List;
+
+/**
+ * A suite of conversions where the time is averaged and then averaged again to give a relatively stable time in milliseconds
+ * that a conversion takes.
+ */
+public class SpeedTest {
```
saquino0827: Now we need PingTest 😆

benchmark/src/main/java/gov/cms/qpp/conversion/SpeedTest.java
```diff
@@ -0,0 +1,73 @@
...
+	 *
+	 * @return The average of the averages.
+	 * @throws InterruptedException When sleeping for one second is interrupted.
+	 */
+	private static Double testSuite() throws InterruptedException {
+		List<Double> testSuite = Lists.newArrayList();
+
+		for (int lcv = 0; lcv < SETS_PER_SUITE; lcv++) {
+			testSuite.add(testSet());
+			Thread.sleep(1000);
```
omnivion: why?

benchmark/src/main/java/gov/cms/qpp/conversion/SpeedTest.java
```diff
@@ -0,0 +1,73 @@
...
+import java.nio.file.Paths;
+import java.util.List;
+
+/**
+ * A suite of conversions where the time is averaged and then averaged again to give a relatively stable time in milliseconds
+ * that a conversion takes.
+ */
+public class SpeedTest {
+	private static final int SETS_PER_SUITE = 10;
+	private static final int TESTS_PER_SET = 20;
```
omnivion: this number is probably not high enough. When running in -server mode, the jvm invocation threshold for method compiling is 10k. In -client it's 1k.

omnivion:
I don't think an improper performance test is something we should pull in as official, supported code.

halprin:
@omnivion I agree, we shouldn't pull in an improper performance test!  What can we do to make this better?
- Create a warm-up that iterates through a conversion around 200 times?  My `java` specifies `Tier3InvocationThreshold` (client) as `200`.
- Tell users to specify `Tier3InvocationThreshold`, `Tier3BackEdgeThreshold`, and `Tier3MinInvocationThreshold` to a very low number, and create a small warm-up that gets us quickly over the threshold?
- Something else?

halprin:
No more love for this PR.  😢

If I am to understand @omnivion, he doesn't like the idea of having this as part of our official code in the repository.  I'm cool with closing this PR without merging, but does anyone else have a thought on this PR, @clydet or @sfradkin?

halprin:
Meh, I'm closing this.

PR #597, opened 2018-01-30
QPPCT-625: prevent spring boot password creation

clydet:
### Information
- QPPCT-625

- prevent logging of generated password

### Checklist 


rest-api/src/main/resources/application.properties
```diff
@@ -1,3 +1,4 @@
 spring.main.banner-mode=off
 spring.http.multipart.max-file-size=-1
 spring.http.multipart.max-request-size=-1
+security.user.password=not_used
```
sfradkin: hah, that was it??

PR #607, opened 2018-02-06
Refactor project plugin sections, add modernizer-maven-plugin

omnivion:
### Information
- JIRA story QPPCT-649.

### Checklist 


commons/pom.xml
```diff
@@ -15,9 +15,35 @@
 	<packaging>jar</packaging>
 
 	<properties>
-		<requiredCodeCoverage>0.95</requiredCodeCoverage>
+		<requiredCodeCoverage>0.05</requiredCodeCoverage>
```
clydet: that bad huh?

omnivion: well, it could have been 0.08, but I'd rather stick to nice roundish numbers

PR #611, opened 2018-02-08
Chore/qppct 635 more validation api boundary tests 2

clydet:
### Information
- QPPCT-635

- validating the correctness of assumptions regarding validation api messages

### Checklist 

### DIFF
--- a/rest-api/src/test/java/gov/cms/qpp/conversion/api/acceptance/ValidationApiFailureAcceptance.java
+++ b/rest-api/src/test/java/gov/cms/qpp/conversion/api/acceptance/ValidationApiFailureAcceptance.java
@@ -14,8 +14,11 @@ import org.jdom2.filter.Filter;
 import org.jdom2.filter.Filters;
 import org.jdom2.xpath.XPathExpression;
 import org.jdom2.xpath.XPathFactory;
-import org.junit.jupiter.api.Disabled;
+import org.junit.jupiter.api.Tag;
 import org.junit.jupiter.api.extension.ExtendWith;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
 import org.springframework.http.HttpStatus;
 
 import java.io.ByteArrayInputStream;
@@ -24,6 +27,7 @@ import java.io.StringWriter;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.function.Consumer;
+import java.util.stream.Stream;
 
 import static com.google.common.truth.Truth.assertThat;
 import static io.restassured.RestAssured.given;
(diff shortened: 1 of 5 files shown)


rest-api/src/test/java/gov/cms/qpp/conversion/api/acceptance/ValidationApiFailureAcceptance.java
```diff
@@ -0,0 +1,148 @@
...
+
+	@AcceptanceTest
+	void testSuccess() {
+		Response response = performRequest(getQrda(null));
+
+		assertThat(response.getStatusCode()).isEqualTo(HttpStatus.CREATED.value());
+	}
+
+	@ParameterizedTest
+	@Tag("acceptance")
```
omnivion: Let's add a meta annotation @Acceptance, and annotate @AcceptanceTest with it

rest-api/src/test/java/gov/cms/qpp/conversion/api/acceptance/ValidationApiFailureAcceptance.java
```diff
@@ -0,0 +1,148 @@
...
+
+		assertThat(response.getStatusCode()).isEqualTo(HttpStatus.UNPROCESSABLE_ENTITY.value());
+		AllErrors blah = response.getBody().as(AllErrors.class);
+		blah.getErrors().stream().flatMap(error -> error.getDetails().stream())
+			.forEach(verifyDetail(comparison, qrda));
+	}
+
+	private static Stream<Arguments> getFailureScenarios() {
+		return Stream.of(
+			Arguments.of("20160101", new HashMap<String, String>() {{
```
omnivion: Could use Collections#singletonMap

rest-api/src/test/java/gov/cms/qpp/conversion/api/acceptance/ValidationApiFailureAcceptance.java
```diff
@@ -0,0 +1,148 @@
...
+		return given()
+			.multiPart("file", "file", qrda.getBytes())
+			.when()
+			.post("/");
+	}
+
+	private String getQrda(Map<String, String> overrides) {
+		return fixture.execute(new StringWriter(), defaults(overrides))
+			.toString()
+			.replaceAll("\n", "");
```
omnivion: .replace should have the same affect, and avoids the regex methods

rest-api/src/test/java/gov/cms/qpp/conversion/api/acceptance/ValidationApiFailureAcceptance.java
halprin: We already have an annotation `AcceptanceTest`, which is used above.  Is the reason that annotation isn't used here is because the `@Test` of `AcceptanceTest` interferes with the `@ParameterizedTest` here?  If so, perhaps make an `AcceptanceParameterizedTest` annotation?

PR #613, opened 2018-02-09
Docker Script Fix

saquino0827:
### Information
- QPPCT-661

- Change docker_deploy.sh to return failure on an ansible failure

### Checklist 
- [n/a] All JUnit tests pass (`mvn clean verify`).

### DIFF
--- a/tools/bin/docker_deploy.sh
+++ b/tools/bin/docker_deploy.sh
@@ -14,8 +14,12 @@ function run_ansible() {
 }
 
 if [[ "$CIRCLE_BRANCH" == "master" || ( ! -z $DOCKER_DEPLOY_OTHER_BRANCH && "$CIRCLE_BRANCH" == "$DOCKER_DEPLOY_OTHER_BRANCH" ) ]]; then
-	build_and_tar_converter_image
+	#build_and_tar_converter_image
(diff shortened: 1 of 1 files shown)


tools/bin/docker_deploy.sh
```diff
@@ -16,6 +16,10 @@ function run_ansible() {
 if [[ "$CIRCLE_BRANCH" == "master" || ( ! -z $DOCKER_DEPLOY_OTHER_BRANCH && "$CIRCLE_BRANCH" == "$DOCKER_DEPLOY_OTHER_BRANCH" ) ]]; then
 	build_and_tar_converter_image
 	run_ansible ${DOCKER_DEPLOY_HOSTS} ${DOCKER_DEPLOY_REMOTE_USERNAME} ${DOCKER_DEPLOY_SUDOER_PASSWORD}
+	if [[ $? -ne 0 ]]; then
```
saquino0827: My spaces aren't incorrect??

saquino0827: Much better

omnivion:
Can you explain the changes a bit more to the less-than-proficient shell scripters (me)?

saquino0827:
@omnivion sure. It is checking ansible's response to see if it is a non-zero response and purposefully exiting with -1. Before, we were getting the failure response from ansible but then just ignoring it and continuing to run.

PR #617, opened 2018-02-13
Bugfix for stratifier UUID insensitivity

saquino0827:
### Information
- QPPCT-670

- Change stratifier uuid validation check to be case insensitive.

### Checklist 


saquino0827:
Please do not delete this branch after merging into `develop`. This branch will be tagged after being reviewed and merged.

sfradkin:
Not to be too picky, but can you add back another stratifier to the fixture that is in all caps so that we cover that case?

saquino0827:
@sfradkin there are many stratifies in there with all upper case id's. I simple changed 2 to have lower cases.

PR #621, opened 2018-02-16
Return amazon errors directly

omnivion:
### Information
- JIRA story QPPCT-664.

### Checklist 


rest-api/src/main/java/gov/cms/qpp/conversion/api/controllers/v1/ExceptionHandlerControllerV1.java
```diff
@@ -82,6 +84,16 @@ public ExceptionHandlerControllerV1(final AuditService auditService) {
 		return new ResponseEntity<>(exception.getMessage(), httpHeaders, HttpStatus.NOT_FOUND);
 	}
 
+	@ExceptionHandler(AmazonS3Exception.class)
+	@ResponseBody
+	ResponseEntity<String> handleGenericS3Exception(AmazonS3Exception exception) {
```
halprin: I guess we could do this, but I was thinking of something simpler.  In `StorageServiceImpl#getFileByLocationId`, if `amazonS3#getObject` throws an exception, catch it and instead `throw new NoFileInDatabaseException(FILE_NOT_FOUND)`.

What do you think of that?

rest-api/src/main/java/gov/cms/qpp/conversion/api/services/StorageServiceImpl.java
```diff
@@ -53,8 +54,9 @@
 		final String bucketName = environment.getProperty(Constants.BUCKET_NAME_ENV_VARIABLE);
 		final String kmsKey = environment.getProperty(Constants.KMS_KEY_ENV_VARIABLE);
 		if (Strings.isNullOrEmpty(bucketName) || Strings.isNullOrEmpty(kmsKey)) {
-			API_LOG.warn("No bucket name is specified or no KMS key specified.");
-			return CompletableFuture.completedFuture("");
+			String error = "No bucket name is specified or no KMS key specified.";
+			API_LOG.warn(error);
+			return FutureHelper.exceptionally(new IllegalStateException(error));
```
halprin: We don't want to do this because that will prevent the audit information in DynamoDB from being written.  This would occur if the the S3 bucket is not specified, but the KMS key and DynamoDB table are specified.

rest-api/src/main/java/gov/cms/qpp/conversion/api/controllers/v1/ExceptionHandlerControllerV1.java
saquino0827: Would that be accurate enough though? Because dynamo and s3 are different even if we use both for storage.

rest-api/src/main/java/gov/cms/qpp/conversion/api/services/AuditServiceImpl.java
```diff
@@ -88,7 +88,7 @@ public AuditServiceImpl(final StorageService storageService, final DbService dbS
 		CompletableFuture<Void> allWrites = CompletableFuture.allOf(
 				storeContent(validationErrorSource).thenAccept(metadata::setConversionErrorLocator),
 				storeContent(qrdaSource).thenAccept(metadata::setSubmissionLocator));
-		return allWrites.whenComplete((nada, thrown) -> persist(metadata, thrown));
+		return allWrites.whenComplete((ignore, thrown) -> persist(metadata, thrown));
```
clydet: Mckayla prefers nada

rest-api/src/main/java/gov/cms/qpp/conversion/api/controllers/v1/ExceptionHandlerControllerV1.java
saquino0827: I think separate the two out would be more accurate imo.

sfradkin: I would like to ensure that we propagate the correct error from S3 and DynamoDB.  If a request to S3 returns a 404, for instance, that 404 should be propagated as the response code here.  I don't want to obscure the downstream error code by wrapping everything in a 400 error.

omnivion: Working on this now @sfradkin

rest-api/src/main/java/gov/cms/qpp/conversion/api/controllers/v1/ExceptionHandlerControllerV1.java
```diff
@@ -98,6 +100,16 @@ public ExceptionHandlerControllerV1(final AuditService auditService) {
 
 		return new ResponseEntity<>(exception.getMessage(), httpHeaders, HttpStatus.NOT_FOUND);
 	}
+
+	@ExceptionHandler(AmazonServiceException.class)
+	@ResponseBody
+	ResponseEntity<String> handleAmazonException(AmazonServiceException exception) {
+		API_LOG.error("An S3 error occured", exception);
```
halprin: Errr, actually this isn't necessarily only an S3 error.  I would say an AWS error occurred.

omnivion: oops forgot to change that, good catch

sfradkin:
Do we have any tests currently for failures as part of the S3/DynamoDB process and how this change affects that process?  We want to be sure that everything still works as it did prior.

PR #622, opened 2018-02-19
Add date filter to CPC+ endpoints

saquino0827:
### Information
- QPPCT-663

- Add date filter to CPC+ endpoint.
- Clean up acceptance tests to start with a clean dynamo db and s3 bucket.

### Checklist 


rest-api/src/test/java/gov/cms/qpp/conversion/api/acceptance/CpcApiAcceptance.java
```diff
@@ -21,18 +35,25 @@
...
 	private static final String CPC_PLUS_PROGRAM_NAME = "CPCPLUS";
-	private static final String NOT_A_CPC_FILE = "beed7ed4-107c-400f-b0de-0c60abc54344";
+	private static AmazonDynamoDB dynamoClient = AmazonDynamoDBClientBuilder.standard().build();
+	private static DynamoDBMapper mapper;
+	private static AWSKMS kmsClient = AWSKMSClientBuilder.defaultClient();
 
 	@BeforeAll
-	static void createUnprocessedItem() {
+	static void setUp() {
+		String kmsKey = "arn:aws:kms:us-east-1:850094054452:key/8b19f7e9-b58c-4b7a-8162-e01809a8a2e9";
```
clydet: This will likely fail for IAM users who aren't privy.
```
$ aws kms encrypt --key-id arn:aws:kms:us-east-1:850094054452:key/8b19f7e9-b58c-4b7a-8162-e01809a8a2e --plaintext meepmawp

An error occurred (AccessDeniedException) when calling the Encrypt operation: 
```

saquino0827: Yes. Unfortunately, we have to set up specific permissions for any user to run that specific kms key.

test-commons/src/main/java/gov/cms/qpp/test/helper/AwsTestHelper.java
```diff
@@ -0,0 +1,53 @@
...
+import com.amazonaws.services.s3.AmazonS3;
+import com.amazonaws.services.s3.AmazonS3ClientBuilder;
+import com.amazonaws.services.s3.model.ObjectListing;
+import com.amazonaws.services.s3.model.S3ObjectSummary;
+import com.google.common.collect.Lists;
+
+import java.util.List;
+import java.util.Map;
+
+public class AwsTestHelper {
```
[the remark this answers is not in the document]
saquino0827: 👍 fixed.

test-commons/src/main/java/gov/cms/qpp/test/helper/AwsTestHelper.java
```diff
@@ -0,0 +1,62 @@
...
+			metadataList.addAll(scanResult.getItems());
+		}
+
+		metadataList.forEach(map -> dynamoClient.deleteItem(TEST_DYNAMO_TABLE_NAME, map));
+	}
+
+	/**
+	 * Cleans up the test s3 bucket by batch for performance purposes
+	 */
+	public static void cleanS3() {
```
omnivion: do we need to implement any delay between batches for aws rate limiting?

saquino0827: I don't believe so because this is strictly an acceptance test dynamodb table and s3 bucket. When I originally ran this we had quite a number of files but now they will always be clean after the tests run. We can implement a delay if needed but I don't think we have high enough load to really need it.

rest-api/src/test/java/gov/cms/qpp/conversion/api/acceptance/CpcApiAcceptance.java
saquino0827: I feel like this can be handled better in the light deployment. Instead of setting up another kms key on our AWS account for local testing.

sfradkin: @saquino0827 @clydet Any traction on a better solution for this?  Or is the solution to just merge this, get the light deployment merged, and then modify afterwards?

rest-api/src/main/java/gov/cms/qpp/conversion/api/model/Constants.java
```diff
@@ -6,6 +6,7 @@
 public class Constants {
 	public static final String DYNAMO_TABLE_NAME_ENV_VARIABLE = "DYNAMO_TABLE_NAME";
 	public static final String KMS_KEY_ENV_VARIABLE = "KMS_KEY";
+	public static final String TEST_KMS_KEY_ENV_VARIABLE = "TEST_KMS_KEY";
```
halprin: I don't think we should put a test based env var in the non-test `Constants` class.  Perhaps make a `TestConstants` class in `test-commons`.

rest-api/src/test/java/gov/cms/qpp/conversion/api/acceptance/CpcApiAcceptance.java
```diff
@@ -48,7 +50,7 @@
 
 	@BeforeAll
 	static void setUp() {
-		String kmsKey = "arn:aws:kms:us-east-1:850094054452:key/8b19f7e9-b58c-4b7a-8162-e01809a8a2e9";
+		String kmsKey = System.getenv(Constants.TEST_KMS_KEY_ENV_VARIABLE);
```
halprin: What about using the `EnvironmentHelper` class and calling `getOrDefault`?

saquino0827: @halprin it's essentially the same thing. I do prefer this as I think it's more clear. I'd just be moving around the Arn.

rest-api/src/main/java/gov/cms/qpp/conversion/api/model/Constants.java
saquino0827: A class that's sole purpose to hold one constant doesn't sound too appealing to me. Maybe we just keep as a constant at the top of `CpcApiAcceptanceTest.java`?

halprin: Good point.  I think moving it to the top of `CpcApiAcceptance.java` is best.

PR #624, opened 2018-02-20
Fix for missing supplemental data codes

saquino0827:
### Information
- QPPCT-671

- Refactored the validation of supplemental data aggregate counts to occur within the same loop as the supplemental codes. This rendered the object that was causing the NPE to not be used anymore.
- Added new tests for each supplemental type to ensure the correct error would output instead of a null pointer exception.

### Checklist 

### DIFF
--- a/converter/src/main/java/gov/cms/qpp/conversion/validate/CpcMeasureDataValidator.java
+++ b/converter/src/main/java/gov/cms/qpp/conversion/validate/CpcMeasureDataValidator.java
@@ -12,7 +12,9 @@ import gov.cms.qpp.conversion.model.validation.MeasureConfig;
 import gov.cms.qpp.conversion.model.validation.MeasureConfigs;
 import gov.cms.qpp.conversion.model.validation.SupplementalData;
 import gov.cms.qpp.conversion.model.validation.SupplementalData.SupplementalType;
+
 import java.util.EnumSet;
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.function.Predicate;
@@ -64,14 +66,24 @@ public class CpcMeasureDataValidator extends NodeValidator {
 		EnumSet<SupplementalData> codes = EnumSet.copyOf(
 				 SupplementalData.getSupplementalDataSetByType(supplementalDataType));
 
-		for (SupplementalData supplementalData : codes) {
-			Node validatedSupplementalNode = filterCorrectNode(supplementalDataNodes, supplementalData);
-
-			if (validatedSupplementalNode == null) {
-				addSupplementalValidationError(node, supplementalData);
+		MeasureConfig config = MeasureConfigs.getConfigurationMap().get(
+			node.getParent().getValue(QualityMeasureIdValidator.MEASURE_ID));
+		String measureId = config.getElectronicMeasureId();
+		if (config == null) {
+			if(measureId != null) {
+				List<String> suggestions = MeasureConfigs.getMeasureSuggestions(measureId);
+				addValidationError(Detail.forErrorAndNode(ErrorCode.MEASURE_GUID_MISSING.format(measureId, suggestions), node));
 			}
+		} else {
+			for (SupplementalData supplementalData : codes) {
+				Node validatedSupplementalNode = filterCorrectNode(supplementalDataNodes, supplementalData);
+
+				if (validatedSupplementalNode == null) {
+					addSupplementalValidationError(node, supplementalData, measureId);
+				}
+			}
+			validateSupplementalDataNodeCounts(node, supplementalDataNodes, measureId);
 		}
-		validateSupplementalDataNodeCounts(node, supplementalDataNodes);
 	}
 
 	/**
@@ -103,17 +115,13 @@ public class CpcMeasureDataValidator extends NodeValidator {
 	 *
 	 * @param node Object being validated
 	 * @param supplementalData Object holding the current code that was validated
+	 * @param measureId current electronic measure identification
 	 */
-	private void addSupplementalValidationError(Node node, SupplementalData supplementalData) {
-		MeasureConfig config =
-				MeasureConfigs.getConfigurationMap()
-						.get(node.getParent().getValue(QualityMeasureIdValidator.MEASURE_ID));
-		if (config != null) {
-			LocalizedError error =
-					ErrorCode.CPC_PLUS_MISSING_SUPPLEMENTAL_CODE.format(supplementalData.getCode(),
-							config.getElectronicMeasureId(), node.getValue(MeasureDataDecoder.MEASURE_TYPE));
-			addValidationError(Detail.forErrorAndNode(error, node));
-		}
+	private void addSupplementalValidationError(Node node, SupplementalData supplementalData, String measureId) {
+		LocalizedError error =
+				ErrorCode.CPC_PLUS_MISSING_SUPPLEMENTAL_CODE_NODE.format(supplementalData.getCode(),
+						measureId, node.getValue(MeasureDataDecoder.MEASURE_TYPE));
+		addValidationError(Detail.forErrorAndNode(error, node));
 	}
 
 	/**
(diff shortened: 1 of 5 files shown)


converter/src/main/java/gov/cms/qpp/conversion/validate/CpcMeasureDataValidator.java
```diff
@@ -64,14 +65,24 @@ private void validateAllSupplementalNodesOfSpecifiedType(
 		EnumSet<SupplementalData> codes = EnumSet.copyOf(
 				 SupplementalData.getSupplementalDataSetByType(supplementalDataType));
 
-		for (SupplementalData supplementalData : codes) {
-			Node validatedSupplementalNode = filterCorrectNode(supplementalDataNodes, supplementalData);
+		String measureId = node.getParent().getValue(QualityMeasureIdValidator.MEASURE_ID);
+		MeasureConfig config = MeasureConfigs.getConfigurationMap().get(measureId);
+		if (config != null) {
+			String electronicMeasureID = config.getElectronicMeasureId();
```
sfradkin: Would it be worthwhile to add some kind of convenience functions on `MeasureConfigs`?  Make it easier to get a `MeasureConfig` without having to reference the map first?  In this code, the `MeasureConfig` isn't used beyond retrieving the `electronicMeasureID`, so having a convenience method there would be helpful and remove a few lines of code here.  There also doesn't appear to be anything going on if `config == null`.  Shouldn't that result in an error condition?

converter/src/main/java/gov/cms/qpp/conversion/validate/CpcMeasureDataValidator.java
```diff
@@ -64,14 +65,24 @@ private void validateAllSupplementalNodesOfSpecifiedType(
...
-			if (validatedSupplementalNode == null) {
-				addSupplementalValidationError(node, supplementalData);
+				if (validatedSupplementalNode == null) {
+					addSupplementalValidationError(node, supplementalData, electronicMeasureID);
+				} else {
+					LocalizedError error = makeIncorrectCountSizeLocalizedError(node, supplementalData.getCode(),
+						electronicMeasureID);
+					check(validatedSupplementalNode)
+						.childMinimum(error, 1, TemplateId.ACI_AGGREGATE_COUNT)
+						.childMaximum(error, 1, TemplateId.ACI_AGGREGATE_COUNT);
```
sfradkin: I've seen this in multiple places... do we not have a validation for asserting that there is a specific count of a `Node`?  Rather than always pairing up `childMinimum` and `childMaximum`?

converter/src/main/java/gov/cms/qpp/conversion/validate/CpcMeasureDataValidator.java
saquino0827: It doesn't because measure configuration is checked and an error is added elsewhere in the code. However, we continue to validate the child nodes. If we don't check to ensure the measure configuration is there, we perform validation with a null measure configuration and get the NPE that we were seeing. I do agree that `MeasureConfig` could be separated into some kind of helper method for it. A couple of other classes use it. It would definitely increase the scope of this task a bit but it can be added here at the cost of probably not finishing at the end of this sprint (which I think is okay). @sfradkin Thoughts?

converter/src/main/java/gov/cms/qpp/conversion/validate/CpcMeasureDataValidator.java
saquino0827: I don't know for sure because I just moved this code. I will double check this. Maybe this is a good chance to refactor `Checker.java`

converter/src/main/java/gov/cms/qpp/conversion/validate/CpcMeasureDataValidator.java
saquino0827: To be honest, I think this refactor should come here because this bug is low priority. I thought because this was a bug, we might want to get this out before the that refactor.

converter/src/main/java/gov/cms/qpp/conversion/validate/CpcMeasureDataValidator.java
saquino0827: Unfortunately, the checker does not have a check to ensure the exact number of children. New checker method for `childExact` incoming.

converter/src/main/java/gov/cms/qpp/conversion/validate/CpcMeasureDataValidator.java
sfradkin: We need to discuss when we want to cut the next release.  If it's a fairly quick refactor, lets get it in here.  Otherwise, if we cut a release before a refactor can be done, we should get the bugfix out.

converter/src/test/java/gov/cms/qpp/conversion/validate/CpcMeasureDataValidatorTest.java
```diff
@@ -135,6 +138,23 @@ void validateFailureSupplementalOtherRaceDataTest() throws Exception {
...
 
+	@Test
+	void validateFailureSupplementalRaceCodeMissing() throws Exception {
+		String failureSexFile = TestHelper.getFixture(MISSING_SUPPLEMENTAL_CODES_FILE);
+		Node placeholder = new QrdaDecoderEngine(new Context()).decode(XmlUtils.stringToDom(failureSexFile));
+		CpcMeasureDataValidator validator = new CpcMeasureDataValidator();
+		Node underTest = placeholder.findFirstNode(TemplateId.MEASURE_DATA_CMS_V2);
+		validator.internalValidateSingleNode(underTest);
+
+		LocalizedError error = ErrorCode.CPC_PLUS_MISSING_SUPPLEMENTAL_CODE.format(SupplementalData.AFRICAN_AMERICAN.getCode(),
```
halprin: Perhaps rename `expectedError`.

converter/src/test/java/gov/cms/qpp/conversion/validate/CpcMeasureDataValidatorTest.java
```diff
@@ -169,6 +189,23 @@ void validateFailureSupplementalFemaleSexDataTest() throws Exception {
...
 
+	@Test
+	void validateFailureSupplementalSexCodeMissing() throws Exception {
+		String failureSexFile = TestHelper.getFixture(MISSING_SUPPLEMENTAL_CODES_FILE);
+		Node placeholder = new QrdaDecoderEngine(new Context()).decode(XmlUtils.stringToDom(failureSexFile));
+		CpcMeasureDataValidator validator = new CpcMeasureDataValidator();
+		Node underTest = placeholder.findFirstNode(TemplateId.MEASURE_DATA_CMS_V2);
+		validator.internalValidateSingleNode(underTest);
+
+		LocalizedError error = ErrorCode.CPC_PLUS_MISSING_SUPPLEMENTAL_CODE.format(SupplementalData.MALE.getCode(),
```
halprin: Perhaps rename `expectedError`.

converter/src/test/java/gov/cms/qpp/conversion/validate/CpcMeasureDataValidatorTest.java
```diff
@@ -203,6 +240,23 @@ void validateFailureSupplementalEthnicityHispanicDataTest() throws Exception {
...
 
+	@Test
+	void validateFailureSupplementalEthnicityCodeMissing() throws Exception {
+		String failureSexFile = TestHelper.getFixture(MISSING_SUPPLEMENTAL_CODES_FILE);
+		Node placeholder = new QrdaDecoderEngine(new Context()).decode(XmlUtils.stringToDom(failureSexFile));
+		CpcMeasureDataValidator validator = new CpcMeasureDataValidator();
+		Node underTest = placeholder.findFirstNode(TemplateId.MEASURE_DATA_CMS_V2);
+		validator.internalValidateSingleNode(underTest);
+
+		LocalizedError error = ErrorCode.CPC_PLUS_MISSING_SUPPLEMENTAL_CODE.format(SupplementalData.NOT_HISPANIC_LATINO.getCode(),
```
halprin: Perhaps rename `expectedError`.

converter/src/test/java/gov/cms/qpp/conversion/validate/CpcMeasureDataValidatorTest.java
```diff
@@ -271,6 +325,23 @@ void validateFailureSupplementalPayerOtherDataTest() throws Exception {
...
 
+	@Test
+	void validateFailureSupplementalPayerCodeMissing() throws Exception {
+		String failureSexFile = TestHelper.getFixture(MISSING_SUPPLEMENTAL_CODES_FILE);
+		Node placeholder = new QrdaDecoderEngine(new Context()).decode(XmlUtils.stringToDom(failureSexFile));
+		CpcMeasureDataValidator validator = new CpcMeasureDataValidator();
+		Node underTest = placeholder.findFirstNode(TemplateId.MEASURE_DATA_CMS_V2);
+		validator.internalValidateSingleNode(underTest);
+
+		LocalizedError error = ErrorCode.CPC_PLUS_MISSING_SUPPLEMENTAL_CODE.format(SupplementalData.MEDICARE.getCode(),
```
halprin: Perhaps rename `expectedError`.

halprin:
Thanks @saquino0827, for changing the variable names.  Unless you get your `Checker` changes in within the next hour, I won't be here to approve this PR.  :(

converter/src/main/java/gov/cms/qpp/conversion/validate/CpcMeasureDataValidator.java
saquino0827: A new story has been added for the MeasureConfigs refactor. It's not that large of a task but this bug is causing issues in production and takes priority.

converter/src/main/java/gov/cms/qpp/conversion/validate/CpcMeasureDataValidator.java
saquino0827: A new story has been made to create a new checker method and refactor all usages of childMin + childMax to use the new method instead.

sfradkin:
Closing this PR as it's superseded by https://github.com/CMSgov/qpp-conversion-tool/pull/628.

The code changes have been promoted as a hotfix instead of part of a sprint release.  Two new stories have already been created (QPPCT-683 and QPPCT-684) for the refactoring discussed in the PR comments.

PR #627, opened 2018-02-22
QPPCT-680: empty property warning

clydet:
### Information
- QPPCT-680

- Added warn logging for empty properties

### Checklist 


commons/src/test/java/gov/cms/qpp/conversion/util/EnvironmentHelperTest.java
```diff
@@ -26,8 +28,11 @@ static void teardown() {
 	@Test
 	void testIsPresentOnRandomString() {
 		String random = UUID.randomUUID().toString();
+		String message = String.format(EnvironmentHelper.NOT_FOUND, random);
+
 		assertWithMessage("Should not have an environment variable with randomized key")
 				.that(EnvironmentHelper.isPresent(random)).isFalse();
+		assertThat(getLogs()).contains(message);
```
omnivion: I think this should be it's own test

commons/src/test/java/gov/cms/qpp/conversion/util/EnvironmentHelperTest.java
```diff
@@ -1,15 +1,17 @@
...
 import java.util.Properties;
 import java.util.UUID;
 
+import gov.cms.qpp.test.logging.LoggerContract;
 import org.junit.jupiter.api.AfterAll;
 import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.Test;
 
-class EnvironmentHelperTest {
+class EnvironmentHelperTest implements LoggerContract {
```
halprin: Sorry, I should have caught this earlier, but we should also have a test for when the env var is present but an empty string.  I believe all the existing tests just use the precondition that the env var doesn't exist.

PR #632, opened 2018-02-27
QPPCT-686: richer missing supplemental data messaging

clydet:
### Information
- QPPCT-686

- Fleshes out error messaging for missing supplemental data

### Checklist 


.circleci/config.yml
```diff
@@ -2,6 +2,7 @@ version: 2
 container_config: &container_config
   docker:
     - image: circleci/openjdk:8-jdk-browsers
+      cmd: ["/bin/bash"]
```
halprin: Just saw that you removed this in the fix build PR.  What's this for?  Does it make all the jobs that use this container configuration execute the commands via bash?

clydet: This was the result of a blog posting from a circleci rep. The guy gave no explanation why it should work just that it "fixed" a problem similar to ours. We plugged it in and notice mixed results so we removed it once we found another approach that worked.

"Does it make all the jobs that use this container configuration execute the commands via bash?" 
Sounds right, but the tasks I see in the build output use bash regardless. So looks like this has no effect. I need to remove this since there's a separate pr for the build fix.

PR #633, opened 2018-02-27
Fix Reporting Parameters Error

saquino0827:
### Information
- QPPCT-662

- Made the `ClinicalDocumentEncoder` ignore any measurement sections that are not ACI, IA, or eCQM sections. This is done here because of a feature of whitelisting rather than blacklisting template id's while decoding.
- While debugging this issue, I found that we still make a qpp.json with a reporting parameters act that is sibling to other measure sections. This causes the submission validation api to treat the reporting parameters act as any other measure section and fail. We may want to discuss if this kind of handholding is a good idea.

### Checklist 


converter/src/main/java/gov/cms/qpp/conversion/encode/ClinicalDocumentEncoder.java
```diff
@@ -90,14 +96,17 @@ private JsonWrapper encodeMeasurementSets(Map<TemplateId, Node> childMapByTempla
...
-			childWrapper = new JsonWrapper();
-			sectionEncoder = encoders.get(child.getType());
-			try {
-				sectionEncoder.encode(childWrapper, child);
-				measurementSetsWrapper.putObject(childWrapper);
-			} catch (NullPointerException exc) {
-				String message = "No encoder for decoder : " + child.getType();
-				throw new EncodeException(message, exc);
+			TemplateId childType = child.getType();
+			if (SECTIONS.contains(childType)) {
```
clydet: 👍

halprin:
@clydet I was thinking we were going to wait until we had our meeting later on today to discuss how we wanted to go about ignoring (if at all) unexpected QRDA3 elements before merging this PR?

PR #637, opened 2018-02-28
Force to use us-east-1 for our acceptance tests

halprin:
### Information
Fixes our acceptance tests in master.

- Forces us to use `us-east-1` with our acceptance tests since CircleCI doesn't know what region to use.  We were already forcing the region with other clients we were making in our acceptance tests, so now we are consistent.

### Checklist 

### DIFF
--- a/test-commons/src/main/java/gov/cms/qpp/test/helper/AwsTestHelper.java
+++ b/test-commons/src/main/java/gov/cms/qpp/test/helper/AwsTestHelper.java
@@ -19,7 +19,7 @@ import java.util.Map;
  * A helper class for aws testing
  */
 public class AwsTestHelper {
-	private static AmazonDynamoDB dynamoClient = AmazonDynamoDBClientBuilder.standard().build();
+	private static AmazonDynamoDB dynamoClient = AmazonDynamoDBClientBuilder.standard().withRegion(Regions.US_EAST_1).build();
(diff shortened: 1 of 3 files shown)


halprin:
Good catch, @omnivion.  I'll fix this up shortly.

halprin:
Fixed, @omnivion!

test-commons/src/main/java/gov/cms/qpp/test/helper/AwsTestHelper.java
```diff
@@ -19,11 +21,25 @@
  * A helper class for aws testing
  */
 public class AwsTestHelper {
-	private static AmazonDynamoDB dynamoClient = AmazonDynamoDBClientBuilder.standard().build();
-	private static AmazonS3 s3Client = AmazonS3ClientBuilder.standard().withRegion(Regions.US_EAST_1).build();
 	public static final String TEST_DYNAMO_TABLE_NAME = "qpp-qrda3converter-acceptance-test";
+
+	private static AmazonDynamoDB dynamoClient = AmazonDynamoDBClientBuilder.standard().withRegion(Regions.US_EAST_1).build();
```
omnivion: why not also final

halprin: Oof, yes, you are correct.  Fixing...  😝

PR #647, opened 2018-03-12
QPPCT-717: Command line doesn't work on Windows

halprin:
### Information
- JIRA story QPPCT-717.

- The regex for determining if a path is correct or not only allowed `/` separators.
- This has been changed to allow a path separator based on the supplied `FileSystem`.

### Checklist 


omnivion:
You could use Jimfs to create the windows file systems for testing.. jimfs-#newFileSystem then pass Configuration#windows

halprin:
@omnivion, the tests actually already do.  `JimfsContract` supplies a `jimfs` method (with one of the `FileSystem`s being Windows), and `@JimfsTest` uses it as a method source.

PR #654, opened 2018-03-16
Data Restoration Script

saquino0827:
### Information
- QPPCT_706

- Add a new script to copy data from one DynamoDb table to another.

### Checklist 


qpp-data-restoration/.gitignore
```diff
@@ -0,0 +1,7 @@
+target/
```
halprin: Doesn't our top-level `.gitignore` file cover all of these?  It should cascade to the sub-directories.

qpp-data-restoration/pom.xml
```diff
@@ -0,0 +1,89 @@
...
+				<groupId>org.apache.maven.plugins</groupId>
+				<artifactId>maven-shade-plugin</artifactId>
+				<version>2.4.1</version>
+				<executions>
+					<execution>
+						<goals>
+							<goal>shade</goal>
+						</goals>
+						<configuration>
+							<dependencyReducedPomLocation>${project.build.directory}/dependency-reduced-pom.xml</dependencyReducedPomLocation>
```
halprin: What is this reduced dependency pom?  It isn't in this PR.

qpp-data-restoration/src/main/java/gov/cms/qpp/conversion/qppdatarestoration/QppDataRestorationApplication.java
```diff
@@ -0,0 +1,79 @@
...
+import java.util.Map;
+import java.util.concurrent.atomic.AtomicInteger;
+
+public class QppDataRestorationApplication {
+	private static final Logger RESTORATION_LOG = LoggerFactory.getLogger(QppDataRestorationApplication.class);
+
+	private static final AmazonDynamoDB DYNAMO_CLIENT =
+		AmazonDynamoDBClientBuilder.standard().withRegion(Regions.US_EAST_1).build();
+
+	public static AmazonDynamoDB getDynamoClient() {
```
halprin: This method is unused.

qpp-data-restoration/src/main/java/gov/cms/qpp/conversion/qppdatarestoration/QppDataRestorationApplication.java
```diff
@@ -0,0 +1,79 @@
...
+import com.amazonaws.services.dynamodbv2.model.ScanRequest;
+import com.amazonaws.services.dynamodbv2.model.ScanResult;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.atomic.AtomicInteger;
+
+public class QppDataRestorationApplication {
```
halprin: If we're going to check this in, we should probably do JavaDocs.

qpp-data-restoration/pom.xml
```diff
@@ -0,0 +1,89 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+
+	<modelVersion>4.0.0</modelVersion>
+
+	<groupId>gov.cms.qpp.conversion</groupId>
```
halprin: Do we want to specify the parent pom?  Not entirely sure.

sfradkin: Agreed, unless we're using something from the parent POM...  I see this as a one-off module that encapsulates a conversion app.

qpp-data-restoration/src/main/java/gov/cms/qpp/conversion/qppdatarestoration/QppDataRestorationApplication.java
```diff
@@ -0,0 +1,79 @@
...
+		AmazonDynamoDBClientBuilder.standard().withRegion(Regions.US_EAST_1).build();
+
+	public static AmazonDynamoDB getDynamoClient() {
+		return DYNAMO_CLIENT;
+	}
+
+	public static void main(String[] args) {
+		if (args.length == 2) {
+			List<Map<String, AttributeValue>> exportTableData = exportData(args[0]);
+			importData(exportTableData, args[1]);
```
sfradkin: We aren't going to run into any memory issues exporting all the data to an in-memory data structure are we?

omnivion: How much data are we looking at?!

qpp-data-restoration/pom.xml
saquino0827: It's generated by the maven shade plugin. It generates a pom without jars included in the regular pom for less duplication when used in other JARS. But ultimately I used this plugin for ease of use and applied the same configuration we had from the converter's pom to this one.

qpp-data-restoration/pom.xml
saquino0827: Let me do a push and see where the code is. I think I forgot to do one more push of tweaks before submitting the PR.

saquino0827: Yeah... the parent pom wasn't in my version lol. That's my fault.

qpp-data-restoration/src/main/java/gov/cms/qpp/conversion/qppdatarestoration/QppDataRestorationApplication.java
saquino0827: If I recall correctly, @halprin said we have about 15,000 or so items. I'll have to ask him to double check but I believe that would be around 20MB of data. I do not think it is of much concern. My 3,000 items test was about 4 MB of data. If necessary, I can change the script to import while exporting. Then, we'd only handle 100 item sized list at a time.

sfradkin: No worries then.

qpp-data-restoration/pom.xml
```diff
@@ -0,0 +1,89 @@
...
+	<artifactId>qpp-data-restoration</artifactId>
+	<version>0.0.1-SNAPSHOT</version>
```
sfradkin: can you make sure the version matches the version in the other POM files?

qpp-data-restoration/src/main/java/gov/cms/qpp/conversion/qppdatarestoration/QppDataRestorationApplication.java
halprin: The entire table is 37.85 MB.

sfradkin: hah, definitely no worries.

qpp-data-restoration/pom.xml
saquino0827: Sure can.

qpp-data-restoration/pom.xml
```diff
@@ -0,0 +1,88 @@
...
+				<version>3.7.0</version>
+				<configuration>
+					<source>${java.version}</source>
+					<target>${java.version}</target>
+				</configuration>
+			</plugin>
+			<plugin>
+				<groupId>org.apache.maven.plugins</groupId>
+				<artifactId>maven-shade-plugin</artifactId>
+				<version>2.4.1</version>
```
omnivion: shade is now 3.1.0. any reason for the old version?

saquino0827: No reason. Updating now 👍

qpp-data-restoration/src/main/java/gov/cms/qpp/conversion/qppdatarestoration/QppDataRestorationApplication.java
```diff
@@ -0,0 +1,83 @@
...
+		});
+		RESTORATION_LOG.info("Finished importing into table " + tableToImportInto);
+	}
+
+	private static void pauseExecution() {
+		try {
+			Thread.sleep(1000);
+		}
+		catch (InterruptedException e) {
+			RESTORATION_LOG.info("Sleep has been interrupted!");
```
omnivion: Maybe "Execution pause was interrupted", and include the exception

PR #662, opened 2018-03-26
Running Docker Compose on Windows

saquino0827:
### Information
- QPPCT-718

- Updates the docker file to be able to run on Docker-toolbox windows 7
    - When I ran docker before the changes, I received the same error the user was getting.
    - Unable to test on windows 10 as docker doesn't work on a windows vm. (Hyper V issues.)
- I excluded our main docker file because I doubt any users would be running our spunk in their docker.

### Checklist 

### DIFF
--- a/DockerfileTest
+++ b/DockerfileTest
@@ -2,6 +2,7 @@ FROM maven:3-jdk-8
 
 RUN mkdir -p /usr/src/app/
 RUN mkdir -p /usr/src/run/
+RUN apt-get install dos2unix
 
 COPY ./ /usr/src/app/
 
@@ -10,6 +11,7 @@ WORKDIR /usr/src/app/
 RUN mvn install -Dmaven.test.skip -Djacoco.skip=true > /dev/null
 RUN cp ./rest-api/target/rest-api.jar /usr/src/run/
 RUN cp -r ./tools/docker/docker-test-artifacts/* /usr/src/run/
+RUN dos2unix /usr/src/run/qppConverterTest.sh
 
 WORKDIR /usr/src/run/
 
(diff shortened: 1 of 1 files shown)


DockerfileTest
```diff
@@ -10,6 +12,7 @@ WORKDIR /usr/src/app/
 RUN mvn install -Dmaven.test.skip -Djacoco.skip=true > /dev/null
 RUN cp ./rest-api/target/rest-api.jar /usr/src/run/
 RUN cp -r ./tools/docker/docker-test-artifacts/* /usr/src/run/
+RUN dos2unix /usr/src/run/qppConverterTest.sh
```
halprin: Wait a second, that file already has LF line endings.  AKA, it already has "UNIX line endings".  Is it possible that the person on Windows is using some weird setting on `git` that forces CRLF line endings?

DockerfileTest
```diff
@@ -2,6 +2,8 @@ FROM maven:3-jdk-8
 
 RUN mkdir -p /usr/src/app/
 RUN mkdir -p /usr/src/run/
+RUN apt-get update && apt-get install
+RUN apt-get install dos2unix
```
halprin: You shouldn't need this separate line.  You can move `dos2unix` to the above line.

DockerfileTest
sfradkin: Even so, maybe it's best to just keep this in here.  At worst it won't change the file.

DockerfileTest
saquino0827: I only separated this for readability.

DockerfileTest
saquino0827: @halprin yes that' what is happening... but I figured we can keep this so the user has little to change when running our docker compose test.

DockerfileTest
halprin: Then you don't need the `apt-get install` on the first line.  Although, I believe the preferred method is to put the `apt-get update` and all installs on the same line.

saquino0827: 👍

PR #670, opened 2018-03-30
QPPCT-748: SimpleUI sends Purpose header

halprin:
### Information
- JIRA story QPPCT-748.

- The SimpleUI sends in the `Purpose` header with `SimpleUI`.
- Reset the default `base href` to `/` and set it to `/developers/conversion-tool-sandbox/` only during deployment.
- Force `allowCredentials` to `true` for CORS since Spring 5.0 defaults to `false`.

### Checklist 

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -104,7 +104,7 @@ jobs:
       - attach_workspace:
           at: /home/circleci
       - run: cd frontend; rm -f yarn.lock && yarn
-      - run: cd frontend; yarn run build;
+      - run: cd frontend; yarn run build --base-href /developers/conversion-tool-sandbox/
       - run: sudo apt-get install python-dev
       - run: curl "https://s3.amazonaws.com/aws-cli/awscli-bundle.zip" -o "awscli-bundle.zip"
       - run: unzip awscli-bundle.zip
@@ -135,7 +135,7 @@ workflows:
       - deploy_api:
           filters:
             branches:
-              only: develop
+              only: story-748-simple_purpose
           requires:
             - checkout
       - integration_test:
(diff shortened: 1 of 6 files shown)


rest-api/src/main/java/gov/cms/qpp/conversion/api/controllers/v1/CpcFileControllerV1.java
```diff
@@ -24,12 +15,21 @@
...
+import java.io.IOException;
+import java.util.List;
+
 /**
  * Controller to handle cpc file data
  */
 @RestController
 @RequestMapping("/cpc")
-@CrossOrigin
+@CrossOrigin(allowCredentials = "true")
```
omnivion: What does this change?

halprin: That allows cookies to be passed cross origin.  In Spring 4.x, if this was unspecified, it defaulted to true.  In Spring 5.x it defaults to false and broke our SimpleUI.  So, we now explicitly say we allow cookies.

.circleci/config.yml
```diff
@@ -104,7 +104,7 @@ jobs:
       - attach_workspace:
           at: /home/circleci
       - run: cd frontend; rm -f yarn.lock && yarn
-      - run: cd frontend; yarn run build;
+      - run: cd frontend; yarn run build --base-href /developers/conversion-tool-sandbox/
```
clydet: are we getting a new route?

halprin: No.  Its the same route as before.  I just moved the base route specification during the deploy instead of hardcoded in our HTML.

PR #673, opened 2018-04-02
Remove Default Decoder

saquino0827:
### Information
- QPPCT-766

- Removes the Default Decoder and tests using it.

### Checklist 


FlexionSonarQube:
SonarQube analysis reported 1 issue

Note: The following issues were found on lines that were not modified in the pull request. Because these issues can't be reported as line comments, they are summarized here:

1. ![MAJOR][MAJOR] [CommandLineRunner.java#L41](https://github.com/CMSgov/qpp-conversion-tool/blob/f01dbcdecb07f22c05098bf66901b331f69602b6/commandline/src/main/java/gov/cms/qpp/conversion/CommandLineRunner.java#L41): Remove this unused "doDefaults" private field. [rule](https://sonarcloud.io/coding_rules#rule_key=squid%3AS1068)

[MAJOR]: https://sonarsource.github.io/sonar-github/severity-major.png 'Severity: MAJOR'

FlexionSonarQube:
SonarQube analysis reported 1 issue

Note: The following issues were found on lines that were not modified in the pull request. Because these issues can't be reported as line comments, they are summarized here:

1. ![MAJOR][MAJOR] [CommandLineRunner.java#L41](https://github.com/CMSgov/qpp-conversion-tool/blob/bcdda3a5b129a17b807f059debfdb0c753f421e2/commandline/src/main/java/gov/cms/qpp/conversion/CommandLineRunner.java#L41): Remove this unused "doDefaults" private field. [rule](https://sonarcloud.io/coding_rules#rule_key=squid%3AS1068)

[MAJOR]: https://sonarsource.github.io/sonar-github/severity-major.png 'Severity: MAJOR'

converter/src/main/java/gov/cms/qpp/conversion/decode/QrdaDecoderEngine.java
```diff
@@ -263,7 +263,7 @@ private QrdaDecoder getDecoder(TemplateId templateId) {
 			}
 
 			Decoder decoder = qppDecoder.getClass().getAnnotation(Decoder.class);
-			TemplateId template = decoder == null ? TemplateId.DEFAULT : decoder.value();
+			TemplateId template = decoder.value();
```
sfradkin: Are we assured that `decoder` will always be non-null here?

saquino0827: I believe I only manually checked this once with a random template ID and did not receive null here. I'll double check to ensure, otherwise. I can make it pass `TemplateId.Unimplemented`

halprin: `decoder` can be `null` only if the `QrdaDecoder` class isn't annotated with `@Decoder`.  Highly unlikely, especially that the decoder wouldn't be in the registry if it wasn't annotated.  But, to be safe, if `decoder` is `null`, we can return `null` from this method.

converter/src/test/java/gov/cms/qpp/conversion/ContextTest.java
```diff
@@ -12,7 +12,6 @@
 
 	@Test
 	void testDoesNotDoDefaultsByDefault() {
```
halprin: We can delete this entire test.

converter/src/test/java/gov/cms/qpp/conversion/decode/QrdaDecoderEngineTest.java
```diff
@@ -327,13 +320,12 @@ void testDontPruneInsignificantChildrenSignificantGrandChildrenWhenDefaults() {
 			QrdaDecoderEngine objectUnderTest = new QrdaDecoderEngine(context);
 			Node decodedNodes = objectUnderTest.decode(rootElement);
 
-			assertNodeCount(decodedNodes, 1, 0, 0);
+			assertNodeCount(decodedNodes, 0, 0, 0);
```
halprin: I believe that this test is now the same as below with the removal of defaults.  @clydet can confirm.

converter/src/test/java/gov/cms/qpp/conversion/ConverterTest.java
```diff
@@ -74,37 +62,6 @@ public void testValidQppStream() {
 		//no exception should be thrown, hence explicitly stating the expected exception is None
 	}
 
-	@Test
-	@PrepareForTest({Converter.class, QrdaValidator.class})
-	public void testValidationErrors() throws Exception {
```
halprin: Unless I am missing something, this test still has merit.  We should change it to use a different `TemplateId` and change `JennyDecoder` to be its own decoder instead of extending from `DefaultDecoder`.

converter/src/test/resources/negative/angerClinicalDocumentValidations.xml
```diff
@@ -224,6 +193,18 @@
 							</item>
 						</list>
 					</text>
+					<entry typeCode="DRIV">
+						<act classCode="ACT" moodCode="EVN">
+							<templateId root="2.16.840.1.113883.10.20.17.3.8"/>
```
halprin: Why did we move the reporting parameters act towards the bottom instead of the top?

converter/src/test/resources/negative/iaSectionContainsWrongChild.xml
```diff
@@ -205,6 +205,7 @@
 							<!-- Implied template Measure Reference templateId -->
 							<templateId root="2.16.840.1.113883.10.20.24.3.98"/>
 							<!-- Improvement Activity Performed Reference and Results templateId -->
+							<templateId root="2.16.840.1.113883.10.20.27.3.3"/>
```
halprin: This'll be a stupid question, but why was this added?

converter/src/test/java/gov/cms/qpp/conversion/ConverterTest.java
saquino0827: Okay. I just flat out removed it as we no longer were using defaults. I can see your point. I'll change this.

converter/src/test/resources/negative/iaSectionContainsWrongChild.xml
saquino0827: This was an accident. Removing.

converter/src/test/resources/negative/angerClinicalDocumentValidations.xml
saquino0827: Reporting parameters was in its own section and not under a specific measure (ACI, IA, ecqm) section. The section itself wasn't being read to show that a reporting parameter start/end dates were missing so I moved it into the correct location to receive the right errors.

converter/src/test/resources/negative/iaSectionContainsWrongChild.xml
saquino0827: Okay. I added this back. The reason for this is because the IA section validator was failing due to not having any incorrect children error. The reason why is because after removing default decoding, the template ID here was not being decoded anymore and thus not being an incorrect child.

converter/src/test/resources/converter/errantDefaultedNode.xml
```diff
@@ -15,7 +15,7 @@
             -->
             <component>
                 <section>
-                    <templateId root="867.5309"/>
+                    <templateId root="2.16.840.1.113883.10.20.27.1.2" extension="2017-07-01"/>
```
halprin: Possibly a problem with this is that we already have the clinical document above.  So the Jenny decoder will be invoked twice.

saquino0827: Hm... I put clinical document so that Jenny decoder would actually go and replace the template id. All the tested passed but you could be right that there is an edge case I missed.

halprin: You could probably just replace it with the ACI section template ID (and update the test to use the ACI section as well).

PR #682, opened 2018-04-10
Update error messages 11 and 12

saquino0827:
### Information
- QPPCT-754

- Title says it all

### Checklist 


commons/src/main/java/gov/cms/qpp/conversion/model/error/ErrorCode.java
```diff
@@ -33,10 +33,9 @@
 			+ "Please review the measures section of your file as it cannot be empty."),
 	AGGREGATE_COUNT_VALUE_NOT_SINGULAR(8, "A single aggregate count value is required"),
 	AGGREGATE_COUNT_VALUE_NOT_INTEGER(9, "Aggregate count value must be an integer"),
-	ACI_MEASURE_PERFORMED_RNR_MEASURE_PERFORMED_EXACT(11, "This ACI Measure Performed RnR requires exactly one "
-		+ "Measure Performed"),
-	ACI_MEASURE_PERFORMED_RNR_MEASURE_ID_NOT_SINGULAR(12, "This ACI Measure Performed RnR's requires a single "
-			+ "Measure ID"),
+	ACI_MEASURE_PERFORMED_RNR_MEASURE_PERFORMED_EXACT(11, "This ACI Reference and Results measure performed is required"),
```
halprin: Wait, that can't be right.  I know that's what it shows in the Jira story, but that error doesn't make sense.

The error mentions the `measure performed` of this `ACI Reference and Results` is required.  _But_ it isn't even there, so you can't reference it with `This`.

If anything, I would say something like `This ACI Reference and Results is missing a required Measure Performed`.  Do others have other ideas?

sfradkin: I like your wording @halprin.

saquino0827: 👍

PR #710, opened 2018-04-19
QPPCT-825: Update deployment version of yarn

halprin:
### Information
- JIRA story QPPCT-825.

- Upgraded to a newer version of CircleCI's Node Docker image.  This uses a new version of Yarn that passes the command line arguments to Angular CLI (`ng`).
- I tested this by running the same commands in our `ui_deployment` CircleCI job associated with building the frontend in a `circleci/node:8.11.1-browsers` Docker container.  I then inspected the output to make sure the base href was set correctly.

### Checklist 

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -81,7 +81,7 @@ jobs:
   ui_test:
     <<: *container_config
     docker:
-      - image: circleci/node:7.10.0-browsers
+      - image: circleci/node:8.11.1-browsers
     steps:
       - attach_workspace:
           at: /home/circleci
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -81,7 +81,7 @@ jobs:
   ui_test:
     <<: *container_config
     docker:
-      - image: circleci/node:7.10.0-browsers
+      - image: circleci/node:8.11.1-browsers
```
omnivion: maybe we should make these versions into anchors

halprin: Good idea!

PR #711, opened 2018-04-20
QPPCT-762: Add better context to messages

halprin:
### Information
- JIRA story QPPCT-762.

- Added another field to `Detail`, the class that gets serialized to report validation errors to our users, called `location`.
  - That field holds a human readable location.
- Gave human readable titles to certain `TemplateId`s.
- When an validation error occurs, the first `Node` or its parent that contains a human readable `TemplateId` is searched for.  The subsequently created `Detail` fills this human readable title as the `location`.
  - If this `Node` is a measure, the measure ID is obtained and appended to the human readable `location`.

### Checklist 

### DIFF
--- a/converter/src/main/java/gov/cms/qpp/conversion/model/TemplateId.java
+++ b/converter/src/main/java/gov/cms/qpp/conversion/model/TemplateId.java
@@ -21,7 +21,7 @@ public enum TemplateId {
 	ACI_NUMERATOR_DENOMINATOR("2.16.840.1.113883.10.20.27.3.28", Extension.JUNE_2017, "ACI Measure"),
 	ACI_NUMERATOR("2.16.840.1.113883.10.20.27.3.31", Extension.SEPTEMBER_2016),
 	ACI_DENOMINATOR("2.16.840.1.113883.10.20.27.3.32", Extension.SEPTEMBER_2016),
-	IA_MEASURE("2.16.840.1.113883.10.20.27.3.33", Extension.SEPTEMBER_2016, "IA Measure"),
+	IA_MEASURE("2.16.840.1.113883.10.20.27.3.33", Extension.SEPTEMBER_2016, "Improvement Activity"),
 	REPORTING_PARAMETERS_ACT("2.16.840.1.113883.10.20.17.3.8"),
 	MEASURE_DATA_CMS_V2("2.16.840.1.113883.10.20.27.3.16", Extension.NOVEMBER_2016),
 	PERFORMANCE_RATE_PROPORTION_MEASURE("2.16.840.1.113883.10.20.27.3.25", Extension.NOVEMBER_2016),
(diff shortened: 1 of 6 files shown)


converter/src/main/java/gov/cms/qpp/conversion/model/TemplateId.java
```diff
@@ -1,31 +1,32 @@
...
 import java.util.Map;
 import java.util.function.Function;
 
-import com.google.common.base.Strings;
-import gov.cms.qpp.conversion.util.EnvironmentHelper;
-
 /**
  * An enumeration of known templates IDs.
  */
 public enum TemplateId {
```
omnivion: What was your process for determining whether the template id has a human readable title?

converter/src/main/java/gov/cms/qpp/conversion/model/error/Detail.java
```diff
@@ -1,31 +1,27 @@
...
+import java.util.Objects;
+
 /**
  * Holds the error information from Validators.
  */
 @JsonIgnoreProperties(ignoreUnknown = true)
 public class Detail implements Serializable {
 	private static final long serialVersionUID = 8818544157552590676L;
 
-	@JsonProperty("errorCode")
```
omnivion: glad we're cleaning this up

converter/src/main/java/gov/cms/qpp/conversion/model/error/Detail.java
```diff
@@ -77,17 +76,35 @@ public static Detail forErrorCode(LocalizedError error) {
...
+		Node importantParentNode = node.findParentNodeWithHumanReadableTemplateId();
+
+		if (importantParentNode != null) {
+			String importantParentTitle = importantParentNode.getType().getHumanReadableTitle();
+			String possibleMeasureId = importantParentNode.getValue("measureId");
+
+			location = importantParentTitle;
+
+			if (!StringUtils.isEmpty(possibleMeasureId)) {
+				location += " " + possibleMeasureId;
```
omnivion: do we want any extra separation here?

converter/src/main/java/gov/cms/qpp/conversion/model/error/Detail.java
```diff
@@ -77,17 +76,35 @@ public static Detail forErrorCode(LocalizedError error) {
 		return detail;
 	}
 
+	private static String computeLocation(Node node) {
+
+		String location = null;
```
omnivion: use a stringbuilder

converter/src/main/java/gov/cms/qpp/conversion/model/TemplateId.java
halprin: The `TemplateId` that has a human readable title will have its title filled into the validation error location.

My thought was when it comes to a human wanting to know the location of an error, a measure will be the deepest that they care about.  If they want more information, they can use the line, column, and XPath information.

converter/src/main/java/gov/cms/qpp/conversion/model/error/Detail.java
halprin: What do you mean by separation?  As in extra lines between the code?

PR #717, opened 2018-05-01
QPPCT-816: reword error message 37

clydet:
### Information
- QPPCT-816

- Rewords error message 37

### Checklist 


ERROR_MESSAGES.md
```diff
@@ -35,7 +35,7 @@ Any text in the following format `(Example)` are considered variables to be fill
 * 34 : CT - Must contain a practice site address for CPC+ conversions
 * 35 : CT - One and only one Alternative Payment Model (APM) Entity Identifier should be specified
 * 36 : CT - CPC+ submissions must contain one Measure section
-* 37 : CT - Must contain correct number of performance rate(s). Correct Number is `(Expected value)`
+* 37 : CT - CPC+ submissions must contain correct number of performance rate(s). Correct Number is `(Expected value)` for measure `(Given measure id)`
```
halprin: Do we want to include the extra ``for measure `(Given measure id)` `` when I just merged PR #711 which adds a reference to the measure in the JSON returned to the user?

clydet: I don't see the harm so long as the value is consistent. Will give #711 a look.

converter/src/main/java/gov/cms/qpp/conversion/util/MeasureConfigHelper.java
```diff
@@ -22,4 +26,29 @@ public static MeasureConfig getMeasureConfig(Node node) {
...
+		return getPrioritizedId(measureConfig);
+	}
+
+	/**
+	 * Find the best available measure id value within the given {@link MeasureConfig}
+	 *
+	 * @param measureConfig a configuration that details a measure
+	 * @return the best available measure id from the given config or a message stating that none were found.
+	 */
+	static String getPrioritizedId(MeasureConfig measureConfig) {
```
halprin: There already exists `MeasureConfigs#getMeasureId` that does something similar to this.  Think you could combine the approaches even though they seem to prioritize a different ID?

clydet: Yeah, but it's not a quick fix. There is some usage of static methods in that helper that would make it difficult to test. Refactor gets a little messy given that the helper depends on Node (in converter) and MeasureConfigs static methods (in common) so it may require some shuffling unless I can find a clean way to unknot the dependencies on static methods. I'll ruminate on it for a bit.

converter/src/test/java/gov/cms/qpp/conversion/util/MeasureConfigHelperTest.java
```diff
@@ -9,13 +12,65 @@
...
 		Node measureNode = new Node(TemplateId.MEASURE_REFERENCE_RESULTS_CMS_V2);
-		measureNode.putValue(MeasureConfigHelper.MEASURE_ID, "40280381-51f0-825b-0152-229bdcab1702");
+		measureNode.putValue(MeasureConfigHelper.MEASURE_ID, THE_UUID);
 		MeasureConfig config = MeasureConfigHelper.getMeasureConfig(measureNode);
 
 		assertThat(config).isNotNull();
 	}
+
+	@Nested
+	@DisplayName("Describe measure value priority")
```
sfradkin: Thanks for adding the @DisplayName annotations

PR #739, opened 2018-05-29
Add CMS Electronic mesure id

saquino0827:
### Information
- QPPCT-856

- Adds the CMS electronic measure id to the location if it is a valid measure id

### Checklist 


converter/src/main/java/gov/cms/qpp/conversion/util/MeasureConfigHelper.java
```diff
@@ -27,6 +27,22 @@ public static MeasureConfig getMeasureConfig(Node node) {
...
 	}
 
+	/**
+	 * Gets the electronic measure id by uuid or defaults to empty if none exists
+	 *
+	 * @param uuid identifier used to fined the electronic measure id
+	 * @return electronic measure id
+	 */
+	public static String getMeasureConfigIdByUuidOrDefault(String uuid) {
+		String electronicMeasureId = "";
```
omnivion: why return a blank? What if config.getElectronicMeasureId() returns null

saquino0827: Just to initialize it `electronicMeasureId` with a fast filler. I changed it to null however.

converter/src/main/java/gov/cms/qpp/conversion/util/MeasureConfigHelper.java
```diff
@@ -27,6 +27,22 @@ public static MeasureConfig getMeasureConfig(Node node) {
...
+	 * Gets the electronic measure id by uuid or defaults to null if none exists
+	 *
+	 * @param uuid identifier used to fined the electronic measure id
+	 * @return electronic measure id
+	 */
+	public static String getMeasureConfigIdByUuidOrDefault(String uuid) {
+		String electronicMeasureId = null;
+		MeasureConfig config = MeasureConfigs.getConfigurationMap().get(uuid);
+		if (config != null) {
+			electronicMeasureId = config.getElectronicMeasureId();
```
omnivion: just return from here instead of creating a var

converter/src/main/java/gov/cms/qpp/conversion/model/error/Detail.java
```diff
@@ -106,6 +107,10 @@ private static String computeLocation(Node node) {
 			if (!StringUtils.isEmpty(possibleMeasureId)) {
 				location.append(" ");
 				location.append(possibleMeasureId);
+				String possibleElectronicMeasureId = MeasureConfigHelper.getMeasureConfigIdByUuidOrDefault(possibleMeasureId);
+				if (!StringUtils.isEmpty(possibleElectronicMeasureId)) {
```
omnivion: Since you’re appending to a string builder, don’t use string concat, instead just call append 3 times

saquino0827: Good catch!

PR #748, opened 2018-07-02
Adds integration profile to build

omnivion:
### Information
- JIRA story QPPCT-854.

### Checklist 


pom.xml
```diff
@@ -189,7 +191,8 @@
 						<trimStackTrace>false</trimStackTrace>
 						<useFile>false</useFile>
 						<properties>
-							<excludeTags>performance</excludeTags>
+							<excludeTags>${junit.excludes}</excludeTags>
+							<includeTags>${junit.includes}</includeTags>
```
saquino0827: With the includes tag having integration and acceptance, is an empty exclude necessary?

omnivion: Yes - we don't want to run the acceptance/integration tests in the normal build, and we need to clear them from excludes even when includes is present.

PR #835, opened 2018-12-28
only copy what's necessary to satisfy Dockerfile

clydet:
### Information
- JIRA 

- added .dockerignore to limit what's copied to container

### Checklist 


omnivion:
sweeeeeet

PR #842, opened 2019-01-03
Warnings cleanup

duselmann:
### Information
- Fixes #?.
- JIRA story ?.

- removed unused imports
- removed unused variables
- excluded junit from secondary projects
- consolidated use of raw generic types
- switched from using hidden internal sun classes to exposed interfaces for future java version compliance.

### Checklist 
- [-] New unit tests written to cover new functionality.
- [-] Added and updated JavaDocs for non-test classes and methods.
- [-] Updated documentation (`README.md`, etc.) depending if the changes require it.

dash '-' in checklist means 'N/A'

This is largely for me to re-familiarize myself with the code base. There are a few risks with this code change that should be considered prior to merging.
1) The generics assumptions I made might be incorrect. Please examine the List<Map> conversions to ensure I made the correct improvements. Furthermore, I could not remove them all because of a legacy library in use. We would have, and probably should, upgrade to a properly generics implementation. The method call requests a generic Class<> rather than an instance. This forces us to send Map.class rather than something like new Map<String, Object>.
2) The dead code removed might be accessed by reflection. There are two instances I had to remove because tests accessed the private members within junit refection calls. The risk here is that if coverage of these sections is second hand (covered by a test for another method or class) then it might be the case that some call is anticipating the private member. Furthermore, if the deleted code setting an unused variable had a consequence outside setting the local variable then it could have adverse effects.
3) I replaced two uses of sun.com impl classes in favor of the exposed w3c.org interfaces. This should be fine but could have consequences that I am unaware. The original programmer might have been working around an issue.


rest-api/src/test/java/gov/cms/qpp/conversion/api/acceptance/ValidationApiFailureAcceptance.java
```diff
@@ -58,21 +55,6 @@ void testBadPerformanceStart(String comparison, Map<String, String> override) {
 			.forEach(verifyDetail(comparison, qrda));
 	}
 
-	private static Stream<Arguments> getFailureScenarios() {
```
omnivion: this is being used by testBadPerformanceStart method

duselmann: Yes, I noticed that when tests failed. I made the appropriate changes and forgot to commit them.

duselmann:
Once this is accepted, I will update with the parent branch (develop) before merging. I would also like to merge this with my work on Sonar issues before (maybe) submitting a PR for that work.

PR #847, opened 2019-01-08
QPPCT-1004 : Pull validation from s3

saquino0827:
### Information
- Adds the s3 portion of the APM to NPI validation. 
- The decision was made to use APM as the key value as it was unique, unlike the NPI value.
- The file is encrypted within an s3 bucket and configured via environment variables.

### Checklist 


omnivion:
Add some unit tests

rest-api/src/main/java/gov/cms/qpp/conversion/api/model/CpcValidationInfo.java
```diff
@@ -0,0 +1,44 @@
...
+
+	@JsonProperty("apm_entity_id")
+	public String getApm() {
+		return apm;
+	}
+
+	public void setApm(final String apm) {
+		this.apm = apm;
+	}
+}
```
duselmann: While this is a POJO, unit test coverage has been added anyhow.

rest-api/src/main/java/gov/cms/qpp/conversion/api/services/QrdaServiceImpl.java
```diff
@@ -39,6 +46,15 @@ public ConversionReport convertQrda3ToQpp(Source source) {
...
 	}
 
+	/**
+	 * Retrieve the CPC+ validation file for the QPP service
+	 *
+	 * @return cpc+ validation file.
+	 */
+	public InputStream getCpcPlusValidationFile() {
+		return storageService.getCpcPlusValidationFile();
+	}
```
duselmann: Usually "get" functions are reserved for simple POJO or Collection actions. This one has a side effect of opening a stream to S3. This is more like an open action. This comment could be said for both this method definition and the storage service definition as well.

rest-api/src/main/java/gov/cms/qpp/conversion/api/services/StorageServiceImpl.java
```diff
@@ -102,6 +102,26 @@ public InputStream getFileByLocationId(String fileLocationId) {
...
+			return null;
+		}
+		API_LOG.info("Retrieving CPC+ validation file from bucket {}", bucketName);
+
+		GetObjectRequest getObjectRequest = new GetObjectRequest(bucketName, key);
+		S3Object s3Object = amazonS3.getObject(getObjectRequest);
+
+		return s3Object.getObjectContent();
+	}
+
```
duselmann: See comment on QrdaServiceImpl.java

rest-api/src/main/java/gov/cms/qpp/conversion/api/services/QrdaServiceImpl.java
saquino0827: I see. Would changing the name to `retrieveS3CpcPlusValidationFile` with some better commenting suffice?

rest-api/src/main/java/gov/cms/qpp/conversion/api/model/CpcValidationInfo.java
saquino0827: Yes. I believe these would have been covered in other unit tests that use this object. Thanks for the help though.

rest-api/src/main/java/gov/cms/qpp/conversion/api/services/QrdaServiceImpl.java
omnivion: yeah just refactor this

saquino0827: It's been refactored.

PR #854, opened 2019-01-14
Metadata holder fixes

omnivion:
- *(Please list what you changed.)*
-
-

### Checklist 


converter/src/main/java/gov/cms/qpp/conversion/Converter.java
```diff
@@ -146,6 +147,13 @@ private JsonWrapper encode() {
 		}
 	}
 
+	private JsonWrapper stripped(JsonWrapper wrapper) {
```
omnivion: I had to extract this to a separate method to shut spotbugs up

PR #861, opened 2019-01-23
QPPCT-1008 JsonWrapper Metadata Refactor

duselmann:
### Information
- JIRA story QPPCT-1008.

- Refactors Metadata out of the JsonWrapper named string map.
- Uses the JsonWrapper recursively to hold object rather than Object type.
-

### Checklist 

### DIFF
--- a/converter/pom.xml
+++ b/converter/pom.xml
@@ -111,5 +111,36 @@
 			<artifactId>cloning</artifactId>
 			<version>1.9.10</version>
 		</dependency>
+
+        <dependency>
+             <groupId>org.junit.jupiter</groupId>
+             <artifactId>junit-jupiter-api</artifactId>
+            <version>${junit.version}</version>
+             <scope>test</scope>
+         </dependency>
+         <dependency>
+             <groupId>org.junit.jupiter</groupId>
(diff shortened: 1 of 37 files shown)


duselmann:
I submitted to allow eyes on it throughout the day. There are some quick followups to come.
I checked that this branch was up to date with develop last night and will merge in new code later today. Also, all tests passed and I "increased" coverage but apparently not enough. Will increase coverage as well.

converter/pom.xml
```diff
@@ -111,5 +111,36 @@
 			<artifactId>cloning</artifactId>
 			<version>1.9.10</version>
 		</dependency>
+
+        <dependency>
```
omnivion: these deps should all be automatically imported from test-commons, right?

duselmann: They do not get pulled in properly with Eclipse JUnit engine. Not Sure why. But yes, this is not a problem with command line maven.

duselmann:
There is a potential design improvement in the Path Correlation to be done (by me). The original impl worked with maps and filtered on metadata_holder. Now that this feature is fully passing all tests, it would be nice to refactor that to walk the new JsonWrapper metadata structure.

PR #862, opened 2019-01-23
Measure Reference and Results Bug

saquino0827:
### Information
- Fix for the `value:true` encoding bug in the measure section reference and results node.
- Forces a whitelist in the measure section v2 encoder as it does not need any other template id's encoded. If other template id's need to be encoded here in the future, they can be added.

### Checklist 


converter/src/main/java/gov/cms/qpp/conversion/encode/QualityMeasureIdEncoder.java
```diff
@@ -181,8 +181,8 @@ private void encodeSubPopulation(Node parentNode, JsonWrapper childWrapper, bool
 		this.encodePerformanceNotMet(childWrapper, parentNode);
 
 		for (Node childNode : parentNode.getChildNodes()) {
-			JsonOutputEncoder measureDataEncoder = encoders.get(childNode.getType());
-			if (null != measureDataEncoder) {
+			if (TemplateId.MEASURE_DATA_CMS_V2.equals(childNode.getType())) {
```
omnivion: you can use ==

PR #867, opened 2019-01-31
Fix CMS347v1

saquino0827:
### Information
- QPPCT-1043
   - Creates a new way for encoding CMS347v1. 
   - The new encoding takes the sum of all subpopulations for each metric type. The reason it was added into the encoding process is because decoding cannot be changed. CMS347v1 requires submission with 3 Subpopulation strata and then to be converted into one with the sum of each subpopulations of same metric type be summed.
   - Adds guids into `measures-data.json` for CMS347v1
-
### Checklist 

### DIFF
--- a/commons/src/main/resources/measures-data.json
+++ b/commons/src/main/resources/measures-data.json
@@ -44044,7 +44044,39 @@
 			"electronicHealthRecord": "https://ecqi.healthit.gov/ecqm/measures/cms347v1"
 		},
 		"cpcPlusGroup": "Other_Measure",
-		"eMeasureUuid": "40280382-5b4d-eebc-015b-8245e0fa06b7"
+		"eMeasureUuid": "40280382-5b4d-eebc-015b-8245e0fa06b7",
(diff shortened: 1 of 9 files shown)


commons/src/main/resources/measures-data.json
```diff
@@ -44044,7 +44044,42 @@
 			"electronicHealthRecord": "https://ecqi.healthit.gov/ecqm/measures/cms347v1"
 		},
 		"cpcPlusGroup": "Other_Measure",
-		"eMeasureUuid": "40280382-5b4d-eebc-015b-8245e0fa06b7"
+		"eMeasureUuid": "40280382-5b4d-eebc-015b-8245e0fa06b7",
```
sfradkin: Will the measures-data changes also be made to the source version of the file?

saquino0827: Yes. I have already made PR to change them. I had to verify with Sarah White that it wouldn't break anything on their end.

PR #913, opened 2019-11-15
Allow all cors

omnivion:
### Information
- JIRA story QPPSF-5378.

- *(Please list what you changed.)*
-
-

### Checklist 


saquino0827:
@omnivion Once this is merged, we should update the web hosted s3 bucket to use dev or imp (probably imp because the vpn restrictions in dev). Also, we may want to consider merging `QPPSF-4883` into this one for the front end changes.

saquino0827:
Follow up to please merge in the frontend changes from branch `QPPSF-4883` to this story.

frontend/src/app/app.component.ts
```diff
@@ -2,7 +2,7 @@ import { Component, OnInit } from '@angular/core';
 import { FileSelectDirective, FileDropDirective, FileUploader } from '../../node_modules/ng2-file-upload/ng2-file-upload';
 import 'rxjs/Rx' ;
 
-const URL  = 'https://qpp.cms.gov/api/submissions/converter';
+const URL  = 'http://localhost:8080/';
```
omnivion: is this accurate @saquino0827

saquino0827: For Local host yes, I feel like this was pointed to dev-pre at some point. Let me check.

saquino0827: The dev-pre URL is `https://preview.qpp.cms.gov/api/submissions/converter`. I believe if we wanted to test this in dev, we may have to get our s3 bucket configured to work through the VPN. And for IMP we would need to include the cookie. Localhost lets you deploy it locally and test.

PR #915, opened 2019-11-20
APM NPI TIN Validation update

saquino0827:
### Information
- QPPSF-5374

- Adds tins to the validation.
- Allows for the same APMs to be validated across multiple tin/npi combinations
- APMs are used as the key value of the APM->TIN/NPI list validation because only one APM is allowed but multiple TIN/NPIs are allowed to be submitted.

### Checklist 


rest-api/src/main/java/gov/cms/qpp/conversion/api/internal/pii/SpecPiiValidator.java
```diff
@@ -21,19 +21,31 @@ public SpecPiiValidator(CpcValidationInfoMap file) {
...
-			} else {
-				if (spec.getApm() != null && !spec.getApm().equalsIgnoreCase(apm)) {
+		List<String> tinList = Arrays.asList(
+			node.getValue(ClinicalDocumentDecoder.TAX_PAYER_IDENTIFICATION_NUMBER).split(","));
+
+		List<CpcValidationInfo> tinNpiCombinationList = file.getApmTinNpiCombination().get(apm);
+		if (tinNpiCombinationList == null){
+			validator.addWarning(Detail.forErrorAndNode(ErrorCode.INCORRECT_API_NPI_COMBINATION, node));
+		} else {
+			for (int index = 0; index < npiList.size(); index++) {
```
omnivion: how large are these lists? if they're very big, let's cache the size

rest-api/src/main/java/gov/cms/qpp/conversion/api/model/CpcValidationInfo.java
```diff
@@ -1,6 +1,7 @@
 package gov.cms.qpp.conversion.api.model;
 
 import com.fasterxml.jackson.annotation.JsonProperty;
+import com.google.common.base.MoreObjects;
```
omnivion: dead import

rest-api/src/main/java/gov/cms/qpp/conversion/api/model/CpcValidationInfoMap.java
```diff
@@ -1,47 +1,61 @@
...
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.function.Function;
-import java.util.stream.Collectors;
 
 public class CpcValidationInfoMap {
 	private static final Logger DEV_LOG = LoggerFactory.getLogger(CpcValidationInfoMap.class);
-	private Map<String, CpcValidationInfo> apmToSpec;
+	private HashMap<String, List<CpcValidationInfo>> apmTinNpiCombinations;
```
omnivion: why hashmap instead of just map?

rest-api/src/main/java/gov/cms/qpp/conversion/api/internal/pii/SpecPiiValidator.java
saquino0827: I think they can get fairly large, are you referring to caching with Integer.valueOf()? Or is the a preferred way for you.

bijujoseph:
@saquino0827 quick question, are we loading the tin-npi file from fms?

saquino0827:
@bijujoseph currently no but this is the test out validation changes in dev and imp using our current method in with encrypted s3. There is another story that @omnivion is working on to change our data storage from encrypted s3 to FMS.

saquino0827:
@bijujoseph I neglected to mention that we do not load the actual TIN/APM/NPI file in dev and imp. We use dummy data.

rest-api/src/main/java/gov/cms/qpp/conversion/api/internal/pii/SpecPiiValidator.java
bijujoseph: @saquino0827  the complexity here can be reduced if we store the CpcValidationInfo using maps. I am responding to this from the question that @omnivion posted below about caching. 
The QrdaService is injected into the RestController (which means it is a singletone). Also the `gov.cms.qpp.conversion.api.services.QrdaServiceImpl#cpcValidationData` is eagerly cached (2 hours). That means everything is in memory. 
I do not know how the incoming data for validation or access parttern is, I am assuming that the flow is something like this  `apm -> tin -> npi` or `apm -> npi -> tin`
Example:
```
//assuming that the access pattern is apm > tin > npi
//CpcValidationInfoMap.java
class CpcValidationInfoMap {
    Map<String, Map<String, List<String>> apmTinNpiCombinations = .....;
    public convertJsonToMapOfLists()...... populate...correctly the map.... 
    
    public boolean isValidRelation(String apm, String tin, String npi) { 
        if (apmTinNpiCombinations[apm] == null) return false;
        if (apmTinNpiCombinations[apm][tin]== null) return false;
        return (apmTinNpiCombinations[apm][tin]).indexOf(npi) > -1;
   }
}
Then in TNAHelper -> tinNpiCombinationExists you can call isValidRelation method
```

saquino0827: @bijujoseph the problem with this is that one tin can have multiple NPI's associated with it. So you actually cannot do a Tin -> NPI map because of a duplicate keys issue.

omnivion: I meant cache the value of list.size() so we're not calling it 1000s of times if it's a huge list

saquino0827: 👍

PR #928, opened 2019-12-12
Dynamo filter date

saquino0827:
### Information
- QueryPage for DynamoDBMapper was not filtering CreateDate properly. Changed to Query and limited each partition to 10 items. There are 32 partitions so this will allow a maximum of 320 results. This was tested with around 300 results and loaded in a timely manner.

### Checklist 


saquino0827:
@KyleApfel It's probably more a 2 pointer, but it did take a me a little longer to find what the actual issue was because I did not realize the DynamoMapper query page did not play well with the creation date.

PR #931, opened 2019-12-16
Fix submission validation testing url

omnivion:
### Information
- JIRA story QPPSF-5491.

### Checklist 


rest-api/pom.xml
```diff
@@ -16,7 +16,7 @@
 	<description>API end-point for the QRDA3 to QPP conversion</description>
 
 	<properties>
-		<requiredCodeCoverage>0.90</requiredCodeCoverage>
```
saquino0827: Are we lowering the overall test coverage for rest api?

saquino0827: @omnivion I'll approve but we'll need a story to bring it back up to 90.

PR #937, opened 2019-12-19
Masked TIN Warning Output

saquino0827:
### Information
- QPPSF-5486

- Masks the TIN # in the warning output for CPC+ submissions.

### Checklist 


omnivion:
Hey, I approved, but just to make sure: tins are 100% numeric, not formatted like they would be on a ssn card, right? eg 321-99-1234

saquino0827:
@omnivion yes the tin's are 100% numeric.

PR #939, opened 2019-12-23
Standing pull request

omnivion:
Passes "mvn clean install"

QPPSF-5490
QPPSF-5491
QPPSF-5531
QPPSF-5578


saquino0827:
Can you add a quick description of the changes in this PR? Also, was `mvn verify` ran on this since we're in the middle of swapping from circle to jenkins? @omnivion

PR #954, opened 2020-01-13
Refactor Error To Problem

saquino0827:
### Information
- QPPSF-5514

- Due to Errors and Warnings being shared, this refactor is to give better details for this. I spoke with @omnivion and he wants to split warnings and errors to not be shared however that is a larger task and I feel should be separated from this first initial change.

### Checklist 


saquino0827:
@omnivion The Json object does get changed from `ErrorCode` to `ProblemCode`. I don't know if that is an issue for the frontend UI. We can always close this PR, up the points of the story, and work on this task for the bigger refactor.

omnivion:
> @omnivion The Json object does get changed from `ErrorCode` to `ProblemCode`. I don't know if that is an issue for the frontend UI. We can always close this PR, up the points of the story, and work on this task for the bigger refactor.

My opinion is to not make any API changes during submission window.

saquino0827:
> > @omnivion The Json object does get changed from `ErrorCode` to `ProblemCode`. I don't know if that is an issue for the frontend UI. We can always close this PR, up the points of the story, and work on this task for the bigger refactor.
> 
> My opinion is to not make any API changes during submission window.

K. I think If I change how it outputs ProblemCode back it should be the same as before.

PR #958, opened 2020-01-21
Dependency updates, cleanup

omnivion:
### Information
- JIRA story QPPSF-5651 QPPSF-5652

- Refactor dependencies
- Update dependencies
- Specifically update jjwt dependency

### Checklist 


saquino0827:
@omnivion there are test failures for the commons package with a `mvn clean package`

saquino0827:
@omnivion lgtm, deploying to dev and testing the JWT changes on dev would be the only way to test if it still works. But I don't think the update would break that.

omnivion:
@saquino0827 can I get an approval

PR #984, opened 2020-08-05
Qppsf 6388 migrate ct deployment

jeremy-page:
### Information
- Migrate CT to Fargate ECS
- JIRA story QPPSF-6388.

- Creates ECS Cluster
- Adds ECS service, task and task definition
- ALB for service
- IAM role & associated policies
- Tags tags tags
- KMS configuration
- README.md

### Checklist 

MVP:
Working running container built from new terraform in dev. Tested by deleting the cluster and recreating it with TF, then calling the root & /health URIs on the ALB. Do *not* run `terraform destroy` on this, it will kill the ECR & it's images so you would have to manually push an image up before it will work again even after running apply.

Deployment steps:
1) Changing the place we push the docker container to the ECR in this TF (we can do this in parallel with the current process and test it throughly)
2) Change the kong route to point at the ALBs created by this TF

TODO:
Update with Dynamo & S3, these are relatively static & some are shared with other apps so I am leaving them for now. Created QPPSF-6774 so we don't forget.


infrastructure/terraform/ssm.tf
```diff
@@ -0,0 +1,556 @@
...
+    sensitivity     = var.sensitivity
+    git-origin      = var.git-origin
+  }
+}
+
+resource "aws_ssm_parameter" "org_name" {
+  name        = "/qppar-sf/${var.environment}/conversion_tool/ORG_NAME"
+  description = "ORG_NAME"
+  type        = "String"
+  value       = "cpc-plus-conversion-tool"
```
saquino0827, line 490: We'll have to see what token organization name they used for this value. It's been changed between environments as well which may break some endpoints that are not required now so we can put this issue  off for a little later.

infrastructure/terraform/ssm.tf
```diff
@@ -0,0 +1,556 @@
...
+    sensitivity     = var.sensitivity
+    git-origin      = var.git-origin
+  }
+}
+
+resource "aws_ssm_parameter" "validation_url" {
+  name        = "/qppar-sf/${var.environment}/conversion_tool/VALIDATION_URL"
+  description = "VALIDATION_URL"
+  type        = "String"
+  value       = "https://preview.qpp.cms.gov/api/submissions/public/validate-submission"
```
saquino0827, line 538: This will need to be changed for prod. From what i've been told, the dev-pre validation api was the only one that worked for testing purposes. This is a heads up for when we are deploying to prod.

infrastructure/deploy/deployConversionTool.groovy
```diff
@@ -0,0 +1,56 @@
...
+
+  environment {
+    CLUSTER_NAME='qppsf-conversion-tool'
+    SERVICE_NAME='conversion-tool'
+    GIT_HASH_TAG="""${sh(returnStdout: true, script: 'git rev-parse HEAD').trim()}"""
+    DEPLOY_TIMEOUT='30'
+  }
+  
+  parameters {
+    choice(name: 'env', choices: ['dev', 'impl', 'prod'], description: 'Which environment do you want to run it in?')
```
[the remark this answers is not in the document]
jeremy-page: Kyle said we did not need a predev

infrastructure/terraform/ssm.tf
jeremy-page: The current values were copied out of the S3 bucket

infrastructure/terraform/ssm.tf
jeremy-page: Plan is to set those manually - the TF will create the SSM entries but I'll need to update them manually because we don't want them checked into git. This is true for all of the "SecureString" settings as well as settings that differ in various environments. I am going to update the README

infrastructure/deploy/deployConversionTool.groovy
```diff
@@ -0,0 +1,56 @@
...
+
+    stage("Deploy") {
+  //    when {
+  //      anyOf {
+  //        branch 'develop'; // dev
+  //        branch 'master'; // prod or impl
+  //        branch 'QPPSF-6388_Migrate_CT_Deployment' ; //feature branch
+  //      }
+   //   }
+
```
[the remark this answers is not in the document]
jeremy-page: This is to actually run a new deploy (as opposed to running the TF to create the environments). Still a WIP but not part of this PR, I should have had it in another branch. It's not used by anything yet

infrastructure/terraform/ssm.tf
saquino0827: Yeah its no problem, but I think the last time they made the testing tokens, they labelled some org names with test in the value.

PR #995, opened 2020-08-12
QPPSF-6652: MIPS APM Submissions

saquino0827:
### Information
- QPPSF-6652

- Adds new functionality for CPC+ APMs to submit MIPS data through a new program name.
  - This involves using the `MIPSAPM` program name, will validate as a CPC+ program name, and then rewrite mips over the program name (Originally this needed two quality sections of the same measure values but a program name of CPC+ and MIPS respectively. This is no longer the case and documentation will be update to avoid confusion).
- Requirement changed to no longer require to have both MIPS and CPC+ quality measure section in one submission.
   - A second submission with `MIPSAPM` program name is required to submit mips data.

### Checklist 


converter/src/main/java/gov/cms/qpp/conversion/encode/ClinicalDocumentEncoder.java
```diff
@@ -114,6 +114,11 @@ private JsonWrapper encodeMeasurementSets(Map<TemplateId, Node> childMapByTempla
 
 				sectionEncoder.encode(childWrapper, child);
 				childWrapper.put("source", "qrda3");
+				if (TemplateId.MEASURE_SECTION_V4.getRoot().equalsIgnoreCase(childType.getRoot())
+					&& ClinicalDocumentDecoder.MIPS_APM.equalsIgnoreCase(
+						currentNode.getValue(ClinicalDocumentDecoder.RAW_PROGRAM_NAME))) {
+					childWrapper.put(ClinicalDocumentDecoder.PROGRAM_NAME, ClinicalDocumentDecoder.MIPS.toLowerCase());
```
shadford, line 120: I think this one needs: `ClinicalDocumentDecoder.MIPS.toLowerCase(Locale.getDefault())` its failing with `[ERROR] Use of non-localized String.toLowerCase() `

saquino0827: I missed this after all the develop branch merges. Thanks for pointing it out. Got it fixed and build runs locally.

PR #1009, opened 2020-09-14
Add SSM env variable

saquino0827:
### Information
- QPPSF-6654: adds the org name for RTI. This value is likely to need to be changed manually as the JWT tokens haven't been created yet. However, we will need this env variable when it is created.

### Checklist 


saquino0827:
Sonar will be taken care of in a separate ticket as it is not reading the code coverage report from Jacoco currently.

saquino0827:
Merging as this is already dev. The Sonar failures will be worked on in a later ticket as there is code coverage that is not being tracked by the scanner.

PR #1010, opened 2020-09-17
QPPSF-6546: APM validation file

saquino0827:
### Information
- Allows the rest api to have a configurable APM validation file. Testing environments can now allow any value we choose to be a proper APM. The validation from the Submission API can specify any test APMs they use and we can configure this to work for CT.

### Checklist 

### DIFF
--- a/rest-api/src/test/resources/test_apm_entity_ids.json
+++ b/rest-api/src/test/resources/test_apm_entity_ids.json
@@ -1,0 +1,1 @@
+["DogCow", "TestApmEntityId"]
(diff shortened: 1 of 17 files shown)


rest-api/src/test/resources/test_apm_entity_ids.json
```diff
@@ -0,0 +1 @@
+["DogCow", "TestApmEntityId"]
```
KyleApfel, line 1:
```suggestion
["A123", "1234-0000"]
```
Would it be better to do something that looks more like an APM entity id here?

KyleApfel (approved):
Looks good.

rest-api/src/test/resources/test_apm_entity_ids.json
saquino0827: We could, as these are just test values. The only effect would be resetting the test files to use these. I can go ahead and do that.

saquino0827:
Admin merging until the Sonar is fixed.

PR #1019, opened 2020-10-14
Fix measure 370 to use the measure id over the CMS identifier

saquino0827:
### Information
- QPPSF-6987

- Fixes the measure 370 bug to allow proper validation.

### Checklist 


KyleApfel (approved):
Looks good.

PR #1053, opened 2021-02-11
QPPSF-7512: MIPSAPM Entity ids

saquino0827:
### Information
- Entity Id now decodes and encodes properly for MIPSAPMs
- Sample file updated to reflect a proper MIPSAPM

### Checklist 


converter/src/main/java/gov/cms/qpp/conversion/decode/ClinicalDocumentDecoder.java
```diff
@@ -72,7 +72,7 @@ protected DecodeResult decode(Element element, Node thisNode) {
 				setNationalProviderIdOnNode(element, thisNode);
 			}
 			if (ENTITY_VIRTUAL_GROUP.equals(entityType)) {
-				setVirtualGroupOnNode(element, thisNode);
+				setApmEntityIdOnNode(element, thisNode, ClinicalDocumentDecoder.ENTITY_ID);
```
KyleApfel, line 75: Naming is weird on these. We have a VG entity, but are setting an APM entity id on that record?

converter/src/main/java/gov/cms/qpp/conversion/encode/ClinicalDocumentEncoder.java
```diff
@@ -64,14 +65,13 @@ private void encodeToplevel(JsonWrapper wrapper, Node thisNode) {
...
-				thisNode.getValue(ClinicalDocumentDecoder.PRACTICE_ID));
+		if (Program.isCpc(thisNode)) {
+			wrapper.put(ClinicalDocumentDecoder.ENTITY_ID, thisNode.getValue(ClinicalDocumentDecoder.PRACTICE_ID));
 		}
 
-		if (ClinicalDocumentDecoder.ENTITY_VIRTUAL_GROUP.equals(entityType)) {
-			wrapper.put(ClinicalDocumentDecoder.ENTITY_ID,
-				thisNode.getValue(ClinicalDocumentDecoder.ENTITY_ID));
+		if (ClinicalDocumentDecoder.ENTITY_VIRTUAL_GROUP.equals(entityType) ||
+			(!Program.isCpc(thisNode) && ClinicalDocumentDecoder.ENTITY_APM.equalsIgnoreCase(entityType))) {
```
KyleApfel, line 73: Is Virtual Groups submitting under the cpc+ program name? I've seen that under Groups, but haven't observed it for VG yet.

KyleApfel (commented):
Added a few comments. Have a few questions on the relationship between VG+APM here.

converter/src/main/java/gov/cms/qpp/conversion/encode/ClinicalDocumentEncoder.java
saquino0827: This is for MIPS APM's to be able to have a practiceId for both virtual groups and MIPS apms as I believe it is required for both.

converter/src/main/java/gov/cms/qpp/conversion/decode/ClinicalDocumentDecoder.java
saquino0827: Yeah, as I understood it, it was the entity id name. I look into setting this to be more generic it's not really an entity id but is a practice id I believe.

saquino0827: @KyleApfel after a bit more research, what's happening is there is an identifier in QRDA-III called a Virtual Group Identifier. it looks something like the following when converted into QPP-JSON: 
`"performanceYear": 2020, "entityType": "virtualGroup", "entityId": "x12345",`
I think when Virtual Group was first introduced to the converter, it was thought that the entityId from APMS and the practice ID from CPC+ submissions are all the same. Their is more underlying code that needs to be re-worked and I think we should address it in a separate ticket as the scope of this task was just to properly encode entity Id's for MIPS APMS.

converter/src/main/java/gov/cms/qpp/conversion/encode/ClinicalDocumentEncoder.java
saquino0827: There is code that uses the same identifier label for APM/CPC+ Practice id's and uses it for virtual groups. It maps the name and value like so 'entityId' : 'virtualGroupIdentifier'. I think because these 3 program types have similar identifiers encoded into QPP format, it got a little mixed up.

converter/src/main/java/gov/cms/qpp/conversion/decode/ClinicalDocumentDecoder.java
saquino0827: I will add a name change from `setApmEntityIdOnNode` to `setEntityIdOnNode` as both virtual group and apm's need to have that entity id field. And then add a separate clean up ticket to separate how it finds the virtual group identifier as opposed to the  apm identifier.

KyleApfel (approved):
Looks good!

shadford (approved):

PR #1086, opened 2021-06-17
QPPSF-7546: Add PI measure validation for specific PI measures

saquino0827:
### Information
- https://jira.cms.gov/browse/QPPSF-7546

- Adds measure validation for PI_HIE_1, PI_HIE_4, and both their respective exclusion measures to be rejected if submitted with PI_HIE_5/

### Checklist 


KyleApfel (approved):
Good refactoring and clean up here. Thanks Sam!

converter/src/test/java/gov/cms/qpp/conversion/validate/PiSectionValidatorTest.java
```diff
@@ -0,0 +1,79 @@
...
+
+import gov.cms.qpp.conversion.model.Node;
+import gov.cms.qpp.conversion.model.TemplateId;
+import gov.cms.qpp.conversion.model.error.Detail;
+import gov.cms.qpp.conversion.model.error.ProblemCode;
+import gov.cms.qpp.conversion.model.error.correspondence.DetailsErrorEquals;
+
+class PiSectionValidatorTest {
+
+	private static final String VALID_ACI_MEASURE = "ACI_EP_1";
```
shadford, line 19: Does this check for the 2017 ACI_EP_1 measure only or does it need to switch between PI_EP_1?

shadford (approved):
Looks Good!

converter/src/test/java/gov/cms/qpp/conversion/validate/PiSectionValidatorTest.java
saquino0827: This is an old test that I was refactoring but it uses old test data. Good catch, as the actual measure name is just test data here but it should be updated as well.

shadford (approved):

PR #1089, opened 2021-07-12
QPPSF-7908 PCF Processing

saquino0827:
### Information
- QPPSF-7908

- Adds PCF conversion currently with 2020 measures data info. (Another ticket will handle updating this)
- Adds unit and integration testing of decoding, validating and encoding data.

### Checklist 


converter/src/main/java/gov/cms/qpp/conversion/decode/ClinicalDocumentDecoder.java
```diff
@@ -30,6 +30,7 @@
 	public static final String ENTITY_TYPE = "entityType";
 	public static final String MIPS_PROGRAM_NAME = "mips";
 	public static final String CPCPLUS_PROGRAM_NAME = "cpcPlus";
+	public static final String PCF = "PCF";
```
KyleApfel, line 33: I thought talking with Michelle she was going to keep this all lowercase. Originally it was case insensitive so any caps was going to be accepted, but we talked her down to all lowercase. That said, I see the other variables not being appended with `_PROGRAM_NAME` are always in caps so might be some context I'm missing...

converter/src/test/java/gov/cms/qpp/acceptance/PcfRoundTripTest.java
```diff
@@ -0,0 +1,62 @@
...
+import java.io.IOException;
+import java.net.URISyntaxException;
+import java.net.URL;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.HashMap;
+
+import static com.google.common.truth.Truth.assertThat;
+
+public class PcfRoundTripTest {
```
KyleApfel, line 24: Reading the code, looks similar to what you were saying during sprint planning. Could you give a quick one-line comment here on the class? Don't think it's super necessary for the methods.

KyleApfel (approved):
As far as I can tell everything looks good! Added one comment that might be due to my lack of domain knowledge, then a small blurb on comments where you think makes sense. There one class I knew how it functions only because of your breakdown during sprint planning, but could do with a comment on those.

converter/src/test/java/gov/cms/qpp/acceptance/PcfRoundTripTest.java
saquino0827: Went ahead and added comments for both negative and positive test cases.

converter/src/main/java/gov/cms/qpp/conversion/decode/ClinicalDocumentDecoder.java
saquino0827: To be safe, I changed the conversion to use lowercasing as Conversion Tool's validation only ensures the value is PCF regardless of case sensitivity.

saquino0827:
> As far as I can tell everything looks good! Added one comment that might be due to my lack of domain knowledge, then a small blurb on comments where you think makes sense. There one class I knew how it functions only because of your breakdown during sprint planning, but could do with a comment on those.

Went ahead and added what you requested and fixed the build. Should be good to go now once this build runs.

KyleApfel (approved):
Thanks Sam!!

PR #1102, opened 2021-08-09
Feature/qppsf 8236

anirudh-addepalli:
### Information
- Fixes QPPSF-8236.
- JIRA story https://jira.cms.gov/browse/QPPSF-8236

### Changes proposed in this PR:adsad
- Changed PCF_REQUIRED_MEASURES to v9

### Checklist 


saquino0827:
Did this measure data file: https://github.com/CMSgov/qpp-conversion-tool/blob/develop/commons/src/main/resources/measures-data.json get updated? I thought it would show that as well in the PR.

saquino0827:
> Did this measure data file: https://github.com/CMSgov/qpp-conversion-tool/blob/develop/commons/src/main/resources/measures-data.json get updated? I thought it would show that as well in the PR.

Did a quick test and it's still converting Y4 measure-data. Seems like that POM xml change isn't pulling the data properly

saquino0827 (changes requested):
See my other comments.

saquino0827 (approved):

PR #1107, opened 2021-08-23
QPPSF-8021: 2021 PCF Validation

pkasarski:
### Information
- JIRA story QPPSF-8021

- CPC+ submissions with PI now cause errors instead of warnings. Older tests were changed to pass. I'm not sure if there is supposed to be backwards-compatibility.
- Added test case files from CMS
- Other requirements for this story were already done: two required measures were updated when `cpcPlusGroup` `Outcome_Measure` was updated in the measures data. The test was updated [here](https://github.com/CMSgov/qpp-conversion-tool/pull/1102/files#diff-06e31ec1e4bc8574ad5ed0635330cb7309d073f9566a8c7409581a1a62f12174). The performance period was [already updated](https://github.com/CMSgov/qpp-conversion-tool/pull/1089/files#diff-c6d5bdcf3b183cc8512134df1d41d9a30d94a5ea935d8e74cd26c4f287dc0aea).

### Checklist 


shadford (approved):
Looks Good!

PR #1270, opened 2022-06-27
QPPSF-10417: Jackson/Spring issue

saquino0827:
### Information
- Fixes a Jackson issue with Spring not accepting the `JsonWrapper.toObject` due to redundancy issues. 

### Checklist 


saquino0827:
@bkkwok 
I noticed there were some lingering properties file changes that are not needed so I removed them. 

The main issue is with the update to spring boot and jackson, Spring had issues with the re-used reference of the name `metadata` in the json format we output for conversion tool. What I did was, instead of structuring our response as an Object that reference's the name Metadata twice within itself (Once for actual json Metadata that I believe is made by Jackson, and once for a Metadata object we construct), I made the whole response into a String and formatted the json using the annotation `@JsonRawValue`.

This is the API documentation for it : https://fasterxml.github.io/jackson-annotations/javadoc/2.5/com/fasterxml/jackson/annotation/JsonRawValue.html

I didn't really find a resolution on Stack Overflow as they all recommended a spring property that did not work and could also cause a redundancy loop.

bkkwok (approved):

PR #1363, opened 2023-08-30
QPPSE-884: Update dependencies for security

saquino0827:
### Information
- bump dependency versions

### Checklist 

### DIFF
--- a/acceptance-tests/pom.xml
+++ b/acceptance-tests/pom.xml
@@ -43,7 +43,7 @@
         <dependency>
             <groupId>org.springframework</groupId>
             <artifactId>spring-web</artifactId>
-            <version>5.2.22.RELEASE</version>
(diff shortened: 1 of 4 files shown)


saquino0827:
The org.springframework:spring-web update to 6.0.0 was excluded for now. The main reasons are that we would need to upgrade to java 17 to use vesion 6.0.0 and that the issue with spring-web is with spring remote using RMI. Spring remote is depreciated in our current version of spring-web and we currently do not use it. Updating to spring-web 6.0.0 removes the depreciated code but I believe dependabot is catching the unused code.

saquino0827:
We will need another ticket to update to Java 17 and once that happens, we can update to spring-web 6.0.0

gerickson808 (approved):

Jpec07 (approved):

PR #1403, opened 2023-12-13
[QPPSE-266]: Participation File Automation

saquino0827:
### Information
- Updates new script to download from s3, format and upload the participation file into s3.

### Checklist 

### DIFF
--- a/upload-part-file.sh
+++ b/upload-part-file.sh
@@ -11,7 +11,9 @@ pip install openpyxl
 pip install simplejson
 
 aws s3 cp s3://${PART_FILE_BUCKET}/${PART_FILE} .
+#chmod +x ./tools/scripts/retrieve-fms-file.py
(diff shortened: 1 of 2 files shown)


upload-part-file.sh
```diff
@@ -10,8 +10,6 @@ export AWS_REGION=${AWS_REGION}
 pip install openpyxl
 pip install simplejson
 
-#chmod +x ./tools/scripts/retrieve-fms-file.py
```
sivaksb, line 13: @saquino0827 could you leave it there, so we can use it when we can hit auth endpoints from codebuild for automation?

saquino0827: K i'm adding it back commented but it will need to upload the participation file into s3 where the rest of the script looks to download and reformat the participation file.

sivaksb (approved):

PR #1465, opened 2024-10-09
QPPSE-2009: Added some logs to better troubleshoot errors

sivaksb:
### Information
- All the PY24 changes have been taken care by the individual tickets.  This is just final make over.  With this, all PY24 changes are complete.

### Checklist 

### DIFF
--- a/converter/src/main/java/gov/cms/qpp/conversion/Converter.java
+++ b/converter/src/main/java/gov/cms/qpp/conversion/Converter.java
@@ -81,7 +81,8 @@ public class Converter {
 			Detail detail = Detail.forProblemCode(ProblemCode.NOT_VALID_XML_DOCUMENT);
 			errors.add(detail);
 		} catch (RuntimeException exception) {
-			DEV_LOG.error(ProblemCode.UNEXPECTED_ERROR.getMessage(), exception);
+			DEV_LOG.error(ProblemCode.UNEXPECTED_ERROR.getMessage());
+			DEV_LOG.error("Problem with transform: " + exception);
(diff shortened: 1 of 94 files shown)


converter/src/main/java/gov/cms/qpp/conversion/Converter.java
```diff
@@ -81,7 +81,8 @@ public JsonWrapper transform() {
 			Detail detail = Detail.forProblemCode(ProblemCode.NOT_VALID_XML_DOCUMENT);
 			errors.add(detail);
 		} catch (RuntimeException exception) {
-			DEV_LOG.error(ProblemCode.UNEXPECTED_ERROR.getMessage(), exception);
+			DEV_LOG.error(ProblemCode.UNEXPECTED_ERROR.getMessage());
+			DEV_LOG.error("Problem with transform: " + exception);
```
Jpec07, line 85: Is there a possibility that PII could wind up in these logs?

sivaksb: The TIN would be type checked and validated separately, where it does not print the exception, rather throw specific error message.  For the other fields, when its not properly formatted or invalid value was provided, this would print the input string for troubleshooting.  So, it should be fine in this case I think.

sivaksb: Please ignore all the commits below, since I was working with Krishna to troubleshoot some other issue, but used this branch, which I should have avoided.

Jpec07 (approved):

gerickson808 (approved):

PR #1477, opened 2024-11-14
QPPA-9685: update snyk finding

chetanmunegowda:
### Information
- Fixes https://jira.cms.gov/browse/QPPA-9685

update version

### Checklist 

### DIFF
--- a/rest-api/pom.xml
+++ b/rest-api/pom.xml
@@ -165,7 +165,7 @@
 		<dependency>
 			<groupId>org.springframework</groupId>
 			<artifactId>spring-webmvc</artifactId>
-			<version>5.3.26</version>
+			<version>6.1.13</version>
 		</dependency>
 
 		<dependency>


rest-api/pom.xml
```diff
@@ -165,7 +165,7 @@
 		<dependency>
 			<groupId>org.springframework</groupId>
 			<artifactId>spring-webmvc</artifactId>
-			<version>5.3.26</version>
+			<version>6.1.14</version>
```
sivaksb, line 168: Since its a major update, the rest api fails to start.  Further research would be needed.

```BeanCreationException: Error creating bean with name 'exceptionHandlerControllerV1': Lookup method resolution failed; nested exception is java.lang.IllegalStateException: Failed to introspect Class [gov.cms.qpp.conversion.api.controllers.v1.ExceptionHandlerControllerV1] from ClassLoader [org.springframework.boot.loader.LaunchedURLClassLoader@6193b845]```

Also, there is a ticket in our backlog for the same - https://jira.cms.gov/browse/QPPSE-2333

sivaksb (changes requested):

